Issue #1, opened 2018-12-03
Minimal project setup


ivansenic:
Create the minimal project setup:

- gradle multip-project build (gradle 5.0 would be great if this is out)
  - since we would have only one project not sure if we need any crazy dependency strategy
  - keep `jmh`, `jococo`, etc
- new project for the agent (`inspectit-oce-agent`?)
  - target the `1.8` for now
  - create the agent entry point, implement both `premain` and `agentmain` (simple sysout) -> see https://docs.oracle.com/javase/7/docs/api/java/lang/instrument/package-summary.html
  - create proper manifest in the build
  - copy from the old the utility to check if dependencies are in `1.8` target
  - what else we would need right away from the old project?
- `.gitignore` that works with intellij (at best copy what's generated from Spring initz, I always found these to be good)
- IntelliJ setup
  - save actions plugin, export configuration -> automatic formatting and import org when saved (Ivan can share configuration)
- create `CONTRIBUTE.MD` and share any needed development environment setup

Confirm:
- agent can be added by `-javaagent` (test versions 8 & 11) 
- agent can be attached to any already running JVM with utility like https://github.com/apangin/jattach (test versions 8 & 11)

mariusoe:
> since we would have only one project not sure if we need any crazy dependency strategy
I think we can (and should) keep this simple. I see no benefit if we create a complicated and fancy dependency strategy

PR #14, opened 2018-12-04
Minimal project setup

JonasKunz:
Pull Request for issue #1 .


.gitignore
```diff
@@ -0,0 +1,7 @@
+#IntelliJ
```
ivansenic: the ignore file is not good, the generated `.iml` files from the IntelliJ are not ignored.. What about if we go with default from the spring initz:

```
.gradle
/build/
!gradle/wrapper/gradle-wrapper.jar

### STS ###
.apt_generated
.classpath
.factorypath
.project
.settings
.springBeans
.sts4-cache

### IntelliJ IDEA ###
.idea
*.iws
*.iml
*.ipr
/out/

### NetBeans ###
/nbproject/private/
/nbbuild/
/dist/
/nbdist/
/.nb-gradle/
```

CONTRIBUTING.MD
```diff
@@ -0,0 +1,3 @@
+# Contributing
+We recommend using [IntelliJ](https://www.jetbrains.com/idea/download/#section=windows) as IDE for contributing. We also recommend installing the Gradle Plugin and the [Save Actions Plugin](https://plugins.jetbrains.com/plugin/7642-save-actions).
```
ivansenic: can you add a sub-section `## IDE` because there will be more things to explain here.. And remove the `#section=windows` from the download link ;)

IntelliJ-config/saveactions_settings.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ivansenic: let's have a config directory for this, so move this to the `config/ide/intellij/saveactions_settings.xml`.. We will surely have more configuration to come.. update the description in the contribute file..

gradle.properties
```diff
@@ -0,0 +1 @@
+releaseVersion=0.1
```
ivansenic: should we add a milestone version here? so `0.1.M1`? what do you think?

inspectit-oce-agent/src/main/resources/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,2 @@
+Agent-Class: rocks.inspectit.oce.agent.main.Agent
```
ivansenic: let's also right now add the:
```
Can-Redefine-Classes: true
Can-Retransform-Classes: true
```
just so that somebody does not break the head later on..

settings.gradle
```diff
@@ -0,0 +1,2 @@
+rootProject.name = "inspectit-oce-root"
```
ivansenic: this should be `inspectit-oce` as that's what root directory will be called once you clone the git repo..

gradle.properties
JonasKunz: Makes sense in my opinion.

settings.gradle
JonasKunz: In the old inspectIt rep owe named the root project "inspectit.root" even though the folder usually would be only "inspectit". I think the name of a Gradle-.Project is not necessarily coupled to the name of its folder. However I agree that it might avoid some confusion when naming it just "inspetit-oce".

Issue #4, opened 2018-12-03
Add open-census and configure metric providers


ivansenic:
- Evaluate where to place the OpenCensus dependencies: both on the bootstrap or only the API on the bootstrap and the Impl on the core?
- configure the metric providers
  - start with Prometheus
  - create `Configuration` class per exporter, activated on f.e. `inspectit.oce.exporter.metrics.prometheus.enabled`
  - all properties for the exporter needs to be in the properties (for example `port` for Prometheus, `endpoint` for Stackdriver)
  - make sure all props are validated (spring)
  - test the properties strategy and adapt the aprach if needed

**Important**: at this point make sure that any metric collected in the programatic way in the monitored application will be visible on our configured exporter.

ivansenic:
> After some research I think that we need to place boththe OpenCensus API and the Implementation in the bootstrap: In the API they look via Class.forName for the implementation:
> https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/tags/Tags.java#L99
> This means that the Implementation needs to be "above" the API in the classloader hierarchy. In addition however the Implementation depends on the API. This means that they both need to be on the bootstrap.

This makes even more need for a solution where all OC dependencies are pushed to our class-loader. We must have a better solution for people that simply don't use the OC lib and just want everything done by `inspectit-oce`. Then they don't have any limitations. Can we implement the switch as part of this ticket then please @JonasKunz ? Maybe this can be part of the `agentArgs`, but not sure how would this play with #19.

PR #15, opened 2018-12-06
closes #2: class loading strategy and bootstrap loader

### LINKED ISSUE #2: Class loading strategy and bootstrap loader
Keep our existing class loading strategy (see https://github.com/inspectIT/inspectIT/blob/master/inspectit.agent.java/src/main/java/rocks/inspectit/agent/java/javaagent/JavaAgent.java), meaning that we have a class loader that will load all our dependencies.

But:
- let's have a package inside of our code f.e. `rocks.inspectit.oce.agent.boostrap` where all classes that needed to be pushed to the system class loader are
- pack those separately into `inspectit-bootstrap.jar` and add this jar to the agent jar
- create a possibility to have other dependencies also loaded by the system class loader (but control this with the build parameter, we could later on release the versions without any bootstrap dependencies so that users can use their own versions and don't depend on us)
- the agent jar should look like:
```
/
  - boostrap/
    - inspectit-bootstrap.jar
    - opecensus-api.jar
    ...
  - libs/
    - spring-context.jar
    ... other dependencies loaded by our CL
  - manifest/
  .. agent classes
```

Let's also make some changes from the original `JavaAgent`:
- don't use `java.util.logging` at all, for now simply collect log messages into some kind of list and optionally print that using `System.out`.. Later on we would push these to the created Spring context where we would have a logger for this. 
- adapt code for loading bootstrap and normal dependencies (the existing code should be optimized, as it was written for java 5)
- update the fix related to the `PlatformClassLoader` in order that it works also for java 10, 11, etc.. Currently this fix is bound to work with java 9 only

A very good reference for the ticket is the OC javaagent project: https://github.com/census-instrumentation/opencensus-java/tree/master/contrib/agent

Add a minimal set of dependencies:
- implementation: `spring-context`, `slf4j` with `logback`
- test: `junit5`, `assertj`

JonasKunz:
Implementation for #2 .


inspectit-oce-bootstrap/build.gradle
```diff
@@ -0,0 +1,14 @@
+dependencies {
+    implementation(
+            'io.opencensus:opencensus-impl:0.18.0'
```
ivansenic: should we have `opencensus-ap`i for now in bootstrap and the `opencensus-impl` in the core? impl in the core should exclude the api and api should be `compileOnly` dependency.. we can also leave this for later and we should check in the #4 what's the bottom line.. if you leave it, then please update description of the #4

inspectit-oce-agent/src/main/java/rocks/inspectit/oce/agent/AgentMain.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.jar.JarFile;
+
+/**
+ * Entry point of the agent.
+ *
+ * @author Jonas Kunz
+ */
+public class AgentMain {
+
+    private static final String INSPECTIT_BOOTSTRAP_JAR_PATH = "/inspectit-oce-bootstrap.jar";
```
ivansenic: I see you did not include the versions in here, if you want it's very simple.. push version to `version.info` file during build and add it to the `inspectit-oce-agent.jar`, then just load this as a resource before initializing the path:

```
private static final String INSPECTIT_BOOTSTRAP_JAR_PATH = "/inspectit-oce-bootstrap-{}.jar";

...

String version = ...
INSPECTIT_BOOTSTRAP_JAR_PATH.format(version)
```

I am not sure if this would help with anything actually but could be clearer.. you could also pass the version to the AgentManager..

inspectit-oce-agent/src/main/java/rocks/inspectit/oce/agent/AgentMain.java
```diff
@@ -0,0 +1,113 @@
...
+
+    public static void agentmain(String agentArgs, Instrumentation inst) {
+        //TODO: currently replacing the agent does not really work as all Agent versions share the same namespace in the same classpath
+        premain(agentArgs, inst);
+    }
+
+    public static void premain(String agentArgs, Instrumentation inst) {
+        try {
+            ClassLoader icl = initializeClasspath(inst);
+            AgentManager.startOrReplaceInspectitCore(icl);
```
ivansenic: this still works right, we are able to reference anything from the bootstrap here or? I was thinking that this `AgentManager` would be loaded by the class loader loading the `AgentMain`.. But this seems not to be the case, I tested it.. but actually, this is also how it was in the old one..

ivansenic: This also represents a problem if we would like to push all dependencies to our class loader. Maybe the easiest is to load the `AgentManager` using reflection from the `icl`, and also invoke this method using reflection? Then you could also remove the dependency to the `inspectit-oce-bootstrap` in this project and we keep it totally separated..

inspectit-oce-agent/src/main/java/rocks/inspectit/oce/agent/AgentMain.java
JonasKunz: Tested it out, as expected it does not work: When attaching the newer agent the "version.info" from the already loaded agent is read, as this one comes first on the classpath..
I'll leave this open for #16 .

PR #17, opened 2018-12-12
Spring

JonasKunz:
Initialized spring and added configuration mechanism.


inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/AgentImpl.java
```diff
@@ -1,16 +1,46 @@
...
 public class AgentImpl implements IAgent {
 
+    private AnnotationConfigApplicationContext ctx;
+
     @Override
-    public void start() {
-        System.out.println("Starting Agent");
+    public void start(String cmdArgs, Instrumentation instrumentation) {
+
+        log.info("Starting Agent...");
```
ivansenic: I would add `inspectit-oce` to the start and showdown messages.. Just to be clear which agent is starting..

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/rocks/inspectit/oce/core/config/DirectoryPropertySource.java
```diff
@@ -0,0 +1,146 @@
...
+                loadPropertiesFile(file);
+            } else if (doesFileHaveEnding(file, YAML_ENDINGS)) {
+                loadYamlFile(file);
+            }
+        });
+        return false;
+    }
+
+    private void loadPropertiesFile(Path file) {
+        try {
```
ivansenic: How about if we use the same code as for `loadYamlFile`, you just need another method in the `ConfigurationUtils` and use `PropertiesPropertiesFactoryBean` instead of `YamlPropertiesFactoryBean` ?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/rocks/inspectit/oce/core/config/DirectoryPropertySource.java
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     * Factory method.
+     * Creates a DirectoryPropertySource and places it as last source in the given container.
+     * @param name the name of the DirectoryPropertySource to create. This is used as basename for the PropertySources generated for each contained file.
+     * @param rootDir the directory to parse for config files
+     * @param container the container to append the PropertySources to
+     * @return
+     */
+    public static DirectoryPropertySource addLast(String name, Path rootDir, MutablePropertySources container) {
+        val dps = new DirectoryPropertySource(name,rootDir,container);
```
ivansenic: is the save actions activated for you? this line is not formatted, can you check quickly, please..

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/rocks/inspectit/oce/core/config/InspectitConfig.java
```diff
@@ -0,0 +1,9 @@
+package rocks.inspectit.oce.core.rocks.inspectit.oce.core.config;
+
+import lombok.Data;
+
+@Data
+public class InspectitConfig {
+
+    private String serviceName;
```
ivansenic: It is very important that we carefully document this class and corresponding configuration file as well

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/rocks/inspectit/oce/core/config/PropertySourcesInitializer.java
```diff
@@ -0,0 +1,46 @@
...
+        addAgentDefaultYaml(propsList);
+    }
+
+    private static void addAgentDefaultYaml(MutablePropertySources propsList) {
+        ClassPathResource defaultYamlResource = new ClassPathResource(DEFAULT_CONFIG_PATH, PropertySourcesInitializer.class.getClassLoader());
+        Properties defaultProps = ConfigurationUtils.readYamlsAsProperties(defaultYamlResource);
+        propsList.addLast(new PropertiesPropertySource(DEFAULT_CONFIG_PROPERTYSOURCE_NAME, defaultProps));
+    }
+
+    private static void addCommandLineConfigDir(MutablePropertySources propsList, String cmdLineArgs) {
```
ivansenic: I am not convinced about this passing of the config dir with command line args..

Can we not have this somehow in the properties, so something like:
```
inspectit:
  config:
     file-config:
        enabled: false
        path: 
        watch: true
```

What I want to say is that there should be this file based config only if a user specifies it.. I guess this is again something you need to pass with the env vars in fact or? because I would like tomorrow to have

* service-config
* kubernetes-config
* etc

I am not sure if there is enough of the abstraction here at the moment.. Or you think that we need to still pass with the agruments as it could be too late? this also goes into the thingy what would we like to allow to be changed dynamically right.. 

Maybe a small discussion about this would be nice :)

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/rocks/inspectit/oce/core/config/DirectoryPropertySource.java
```diff
@@ -0,0 +1,146 @@
+package rocks.inspectit.oce.core.rocks.inspectit.oce.core.config;
```
ivansenic: should we move everything related to the property sources to the  `config.propertysource` sub-package just to organize a bit..

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/rocks/inspectit/oce/core/config/ConfigurationCenter.java
```diff
@@ -0,0 +1,36 @@
+package rocks.inspectit.oce.core.rocks.inspectit.oce.core.config;
```
ivansenic: somehow the directories and packages here are wrong for all the files in the `core`? can you check pls what went wrong?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/rocks/inspectit/oce/core/config/InspectitConfig.java
```diff
@@ -0,0 +1,9 @@
+package rocks.inspectit.oce.core.rocks.inspectit.oce.core.config;
```
ivansenic: move to `config.model`

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/rocks/inspectit/oce/core/config/InspectitConfig.java
```diff
@@ -0,0 +1,9 @@
...
+
+import lombok.Data;
+
+@Data
```
ivansenic: we should have protected setters imo?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/PropertySourcesInitializer.java
```diff
@@ -0,0 +1,76 @@
...
+    private static void addAgentDefaultYaml(MutablePropertySources propsList) {
+        ClassPathResource defaultYamlResource = new ClassPathResource(DEFAULT_CONFIG_PATH, PropertySourcesInitializer.class.getClassLoader());
+        Properties defaultProps = PropertyFileUtils.readYamlFiles(defaultYamlResource);
+        propsList.addLast(new PropertiesPropertySource(DEFAULT_CONFIG_PROPERTYSOURCE_NAME, defaultProps));
+    }
+
+    private static void addFileBasedConfiguration(MutablePropertySources propsList, InspectitConfig currentConfig) {
+        String path = currentConfig.getConfig().getFileBased().getPath();
+        Boolean enabled = currentConfig.getConfig().getFileBased().isEnabled();
+        boolean fileBasedConfigEnabled = enabled && path != null && !path.isEmpty();
```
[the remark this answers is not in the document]
JonasKunz: I would not do the check in teh Constructor of DirectoryPropertySource. There is nothing meaningful today in the Constructor if the path exists.

PR #20, opened 2018-12-13
closes #3: Added directory watcher, moved configuration management

### LINKED ISSUE #3: Initialize Spring context and properties
- initialize the spring context on the agent start-up (log messages, add `Instrumentation` to the context we will need this later on)
- create a default properties file
- create properties strategy (we need to brainstorm this a while); idea is that we have set of default properties that would be overwritten with the env vars, but we need a long-term idea how can we deal with external properties. One idea could be to try to load the external files and maintain them together with our default properties. Then the system could always answer to the "property value request" as `if not provided externally then default`. 

Before starting, have a look on:
- https://docs.spring.io/spring/docs/current/spring-framework-reference/
- https://github.com/spring-projects/spring-framework/wiki/Upgrading-to-Spring-Framework-5.x

JonasKunz:


inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/InspectitEnvironment.java
```diff
@@ -0,0 +1,136 @@
...
+    private static final String DEFAULT_CONFIG_PROPERTYSOURCE_NAME = "inspectitDefaults";
+
+
+    /**
+     * Sorted list of all configuration sources.
+     * They are loaded in the given order. The earlier a configuration appears in this list, the higher its priority.
+     * This means that configurations loaded from items appearing earlier in the list overwrite configurations from items appearing later in the list.
+     * In contrast items appearing first in the list can provide information for loading items appearing later in the list.
+     */
+    private static final List<BiConsumer<MutablePropertySources, InspectitConfig>> configurationInitializationSteps = Arrays.asList(
```
ivansenic: let's name this car in a `static final` way `CONFIGURATION_INIT...`

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/InspectitEnvironment.java
```diff
@@ -0,0 +1,136 @@
...
+    /**
+     * The currently active inspectIT configuration.
+     */
+    @Getter
+    private InspectitConfig currentConfig;
+
+    /**
+     * Event drain for publishing {@link InspectitConfigChangedEvent}s.
+     */
+    private ApplicationEventPublisher eventDrain;
```
ivansenic: also `final`?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/InspectitEnvironment.java
```diff
@@ -0,0 +1,136 @@
...
+    private static void addAgentDefaultYaml(MutablePropertySources propsList) {
+        ClassPathResource defaultYamlResource = new ClassPathResource(DEFAULT_CONFIG_PATH, InspectitEnvironment.class.getClassLoader());
+        Properties defaultProps = PropertyFileUtils.readYamlFiles(defaultYamlResource);
+        propsList.addLast(new PropertiesPropertySource(DEFAULT_CONFIG_PROPERTYSOURCE_NAME, defaultProps));
+    }
+
+    private static void addFileBasedConfiguration(MutablePropertySources propsList, InspectitConfig currentConfig) {
+        String path = currentConfig.getConfig().getFileBased().getPath();
+        Path dirPath = Paths.get(path);
+        Boolean enabled = currentConfig.getConfig().getFileBased().isEnabled();
```
ivansenic: Is this really an object Boolean? Can this ever be `null` or what's the deal here?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/filebased/ConfigurationDirectoriesWatcher.java
```diff
@@ -0,0 +1,112 @@
...
+                ws.close();
+            } catch (IOException e) {
+                log.error("Unable to close watch service", e);
+                return;
+            }
+        }
+    }
+
+    private void startWatcherThread() {
+        val watchingThread = new Thread(() -> {
```
ivansenic: OK I get what you did, what I would try to avoid is uncontrolled creation of threads.. A much better way would be the `Executor` usage.. Now the Java8 does bring some default executor stuff, but this one is not a `ScheduledExecutorService`.. As we wound anyway need some scheduling for metrics collections, let's already prepare a field for that.. I see two options:

1. Use Spring scheduling stuff, so just have a `@Scheduled` annotation method (needs @EnableScheduling in the configuration). This approach also makes it easer to test later on..
2. Create our own Scheduled executor which you can auto-wire and use when needed.. Then you can submit runnable or callables and get back a Cancelable object to control if it has to be canceled..

What do you think?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/service/ActivationConfigConditionMet.java
```diff
@@ -0,0 +1,33 @@
...
+    public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata) {
+        InspectitEnvironment env = (InspectitEnvironment) context.getEnvironment();
+        val attribs = metadata.getAnnotationAttributes(ActivationConfigCondition.class.getName());
+        if (attribs == null || attribs.isEmpty()) {
+            throw new RuntimeException("Annotation ActivationConfigCondition is missing!");
+        }
+        InspectitConfig currentConfig = env.getCurrentConfig();
+
+        ExpressionParser parser = new SpelExpressionParser();
+        Expression exp = parser.parseExpression((String) attribs.get("value"));
```
ivansenic: there has to be a way that attribs is an instance of `ActivationConfigCondition`, then you would not need a cast here? maybe it's because of this `val`?

PR #22, opened 2018-12-17
closes #4: Added Prometheus exporter and config validation

### LINKED ISSUE #4: Add open-census and configure metric providers

JonasKunz:
- extended the Agent with a base class for services which can be dynamically activated, deactivated and configured through the inspectIT configuration mechanism
- the prometheus OpenCensus exporter was added as a dynamic service (can be enabled, disabled and reconfigured at runtime)
- added config validation via the hibernate implementation
- moved the package rocks.inspectit.oce.config.service to rocks.inspectit.oce.service

### DIFF
--- a/inspectit-oce-core/build.gradle
+++ b/inspectit-oce-core/build.gradle
@@ -58,7 +58,8 @@ dependencies {
     testImplementation(
             'org.springframework:spring-test:5.1.3.RELEASE',
             'org.apache.httpcomponents:httpclient:4.5.6',
-            'org.mockito:mockito-core:2.23.4'
+            'org.mockito:mockito-core:2.23.4',
+            'org.assertj:assertj-core:3.11.1'
     )
 
 }
--- a/inspectit-oce-core/build.gradle
+++ b/inspectit-oce-core/build.gradle
@@ -1,4 +1,3 @@
-
 plugins {
     id "me.champeau.gradle.jmh" version "0.4.7"
 }
@@ -43,9 +42,21 @@ dependencies {
     implementation(
             'org.springframework.boot:spring-boot:2.1.1.RELEASE',
             'org.yaml:snakeyaml:1.23',
+
+            //data validation
             'javax.validation:validation-api:2.0.0.Final',
-            'ch.qos.logback:logback-classic:1.2.3'
+            'org.hibernate.validator:hibernate-validator:6.0.2.Final',
+            'org.hibernate.validator:hibernate-validator-annotation-processor:6.0.2.Final',
+            'javax.el:javax.el-api:3.0.0',
+            'org.glassfish.web:javax.el:2.2.6',
+
+            'ch.qos.logback:logback-classic:1.2.3',
+            'javax.annotation:javax.annotation-api:1.3.2', //Required for @PostConstruct and @PreDestroy to work in Java9+
+            //OpenCensus exporters
+            "io.opencensus:opencensus-exporter-stats-prometheus:${openCensusVersion}",
+            "io.prometheus:simpleclient_httpserver:${prometheusClientVersion}"
     )
+
 }
 
 task buildOpencensusFatJar(type: Jar) {
--- a/inspectit-oce-core/build.gradle
+++ b/inspectit-oce-core/build.gradle
@@ -44,11 +44,8 @@ dependencies {
             'org.yaml:snakeyaml:1.23',
 
             //data validation
-            'javax.validation:validation-api:2.0.0.Final',
-            'org.hibernate.validator:hibernate-validator:6.0.2.Final',
-            'org.hibernate.validator:hibernate-validator-annotation-processor:6.0.2.Final',
-            'javax.el:javax.el-api:3.0.0',
-            'org.glassfish.web:javax.el:2.2.6',
+            'org.hibernate.validator:hibernate-validator:6.0.13.Final',
+            'org.apache.tomcat.embed:tomcat-embed-el:9.0.13',
 
             'ch.qos.logback:logback-classic:1.2.3',
             'javax.annotation:javax.annotation-api:1.3.2', //Required for @PostConstruct and @PreDestroy to work in Java9+
--- a/inspectit-oce-core/build.gradle
+++ b/inspectit-oce-core/build.gradle
@@ -27,6 +27,7 @@ jmh {
 
 configurations{
     opencensus
+    testImplementation.extendsFrom opencensus
 }
 
 dependencies {
@@ -52,6 +53,12 @@ dependencies {
             //OpenCensus exporters
             "io.opencensus:opencensus-exporter-stats-prometheus:${openCensusVersion}",
             "io.prometheus:simpleclient_httpserver:${prometheusClientVersion}"
+    )
+
+    testImplementation(
+            'org.springframework:spring-test:5.1.3.RELEASE',
+            'org.apache.httpcomponents:httpclient:4.5.6',
+            'org.mockito:mockito-core:2.23.4'
     )
 
 }
--- a/inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/InspectitEnvironment.java
+++ b/inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/InspectitEnvironment.java
@@ -14,6 +14,9 @@ import rocks.inspectit.oce.core.config.filebased.DirectoryPropertySource;
 import rocks.inspectit.oce.core.config.filebased.PropertyFileUtils;
 import rocks.inspectit.oce.core.config.model.InspectitConfig;
 
+import javax.validation.ConstraintViolation;
+import javax.validation.Validation;
+import javax.validation.Validator;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -57,6 +60,11 @@ public class InspectitEnvironment extends StandardEnvironment {
      */
     private final ApplicationEventPublisher eventDrain;
 
+    /**
+     * Validator used for validating configurations.
+     */
+    private Validator validator;
+
     public InspectitEnvironment(ApplicationEventPublisher eventDrain) {
         this.eventDrain = eventDrain;
     }
@@ -91,7 +99,24 @@ public class InspectitEnvironment extends StandardEnvironment {
     }
 
     private void reloadConfig() {
-        currentConfig = Binder.get(this).bind("inspectit", InspectitConfig.class).get();
+        InspectitConfig newConfig = Binder.get(this).bind("inspectit", InspectitConfig.class).get();
+        Validator validator = getValidator();
+        val violations = validator.validate(newConfig);
+        if (violations.isEmpty()) {
+            currentConfig = newConfig;
+        } else {
+            log.error("The given configuration is not valid! The currently active configuration will be kept.");
+            for (ConstraintViolation<InspectitConfig> vio : violations) {
+                log.error("{} (={}) => {}", CaseUtils.camelCaseToKebabCase(vio.getPropertyPath().toString()), vio.getInvalidValue(), vio.getMessage());
+            }
+        }
+    }
+
+    private Validator getValidator() {
+        if (validator == null) {
+            validator = Validation.buildDefaultValidatorFactory().getValidator();
+        }
+        return validator;
     }
 
     @Override
--- a/inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/exporter/PrometheusExporterService.java
+++ b/inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/exporter/PrometheusExporterService.java
@@ -22,12 +22,12 @@ public class PrometheusExporterService extends DynamicallyActivatableService {
     private HTTPServer prometheusClient = null;
 
     public PrometheusExporterService() {
-        super("exporters.metrics.prometheus");
+        super("exporters.metrics.prometheus", "metrics.enabled");
     }
 
     @Override
     protected boolean checkEnabledForConfig(InspectitConfig conf) {
-        return conf.getExporters().getMetrics().getPrometheus().isEnabled();
+        return conf.getExporters().getMetrics().getPrometheus().isEnabled() && conf.getMetrics().isEnabled();
     }
 
     @Override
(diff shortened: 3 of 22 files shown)


inspectit-oce-core/build.gradle
```diff
@@ -43,9 +42,21 @@ dependencies {
     implementation(
             'org.springframework.boot:spring-boot:2.1.1.RELEASE',
             'org.yaml:snakeyaml:1.23',
+
+            //data validation
             'javax.validation:validation-api:2.0.0.Final',
```
ivansenic: this one is not included as part of `org.hibernate.validator:hibernate-validator:6.0.2.Final`

inspectit-oce-core/build.gradle
```diff
@@ -43,9 +42,21 @@ dependencies {
...
-            'ch.qos.logback:logback-classic:1.2.3'
+            'org.hibernate.validator:hibernate-validator:6.0.2.Final',
+            'org.hibernate.validator:hibernate-validator-annotation-processor:6.0.2.Final',
+            'javax.el:javax.el-api:3.0.0',
+            'org.glassfish.web:javax.el:2.2.6',
```
ivansenic: this one and previous one are not really compatible:

```
+--- javax.el:javax.el-api:3.0.0
+--- org.glassfish.web:javax.el:2.2.6
|    \--- javax.el:javax.el-api:2.2.5 -> 3.0.0
```

ivansenic: I checked not what would be the dependencies of the `org.springframework.boot:spring-boot-starter-validation`, are you sure we are actually needing all those `el` stuff? shouldn't this be enough somehow?

```
compileClasspath - Compile classpath for source set 'main'.
\--- org.springframework.boot:spring-boot-starter-validation -> 2.1.1.RELEASE
     +--- org.springframework.boot:spring-boot-starter:2.1.1.RELEASE
     |    +--- org.springframework.boot:spring-boot:2.1.1.RELEASE
     |    |    +--- org.springframework:spring-core:5.1.3.RELEASE
     |    |    |    \--- org.springframework:spring-jcl:5.1.3.RELEASE
     |    |    \--- org.springframework:spring-context:5.1.3.RELEASE
     |    |         +--- org.springframework:spring-aop:5.1.3.RELEASE
     |    |         |    +--- org.springframework:spring-beans:5.1.3.RELEASE
     |    |         |    |    \--- org.springframework:spring-core:5.1.3.RELEASE (*)
     |    |         |    \--- org.springframework:spring-core:5.1.3.RELEASE (*)
     |    |         +--- org.springframework:spring-beans:5.1.3.RELEASE (*)
     |    |         +--- org.springframework:spring-core:5.1.3.RELEASE (*)
     |    |         \--- org.springframework:spring-expression:5.1.3.RELEASE
     |    |              \--- org.springframework:spring-core:5.1.3.RELEASE (*)
     |    +--- org.springframework.boot:spring-boot-autoconfigure:2.1.1.RELEASE
     |    |    \--- org.springframework.boot:spring-boot:2.1.1.RELEASE (*)
     |    +--- org.springframework.boot:spring-boot-starter-logging:2.1.1.RELEASE
     |    |    +--- ch.qos.logback:logback-classic:1.2.3
     |    |    |    +--- ch.qos.logback:logback-core:1.2.3
     |    |    |    \--- org.slf4j:slf4j-api:1.7.25
     |    |    +--- org.apache.logging.log4j:log4j-to-slf4j:2.11.1
     |    |    |    +--- org.slf4j:slf4j-api:1.7.25
     |    |    |    \--- org.apache.logging.log4j:log4j-api:2.11.1
     |    |    \--- org.slf4j:jul-to-slf4j:1.7.25
     |    |         \--- org.slf4j:slf4j-api:1.7.25
     |    +--- javax.annotation:javax.annotation-api:1.3.2
     |    +--- org.springframework:spring-core:5.1.3.RELEASE (*)
     |    \--- org.yaml:snakeyaml:1.23
     +--- org.apache.tomcat.embed:tomcat-embed-el:9.0.13
     \--- org.hibernate.validator:hibernate-validator:6.0.13.Final
          +--- javax.validation:validation-api:2.0.1.Final
          +--- org.jboss.logging:jboss-logging:3.3.2.Final
          \--- com.fasterxml:classmate:1.3.4 -> 1.4.0
```

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/InspectitEnvironment.java
```diff
@@ -91,7 +99,24 @@ public synchronized void readPropertySources(Consumer<PropertySources> propertie
...
+            log.error("The given configuration is not valid! The currently active configuration will be kept.");
+            for (ConstraintViolation<InspectitConfig> vio : violations) {
+                log.error("{} (={}) => {}", CaseUtils.camelCaseToKebabCase(vio.getPropertyPath().toString()), vio.getInvalidValue(), vio.getMessage());
+            }
+        }
+    }
+
+    private Validator getValidator() {
+        if (validator == null) {
+            validator = Validation.buildDefaultValidatorFactory().getValidator();
```
ivansenic: you are not setting the validation instance to `this`?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/exporter/PrometheusExporterService.java
```diff
@@ -0,0 +1,59 @@
...
+ * Can be dynamically started and stopped using the exporters.metrics.prometheus.enabled configuration.
+ */
+@Component
+@Slf4j
+public class PrometheusExporterService extends DynamicallyActivatableService {
+
+    private HTTPServer prometheusClient = null;
+
+    public PrometheusExporterService() {
+        super("exporters.metrics.prometheus");
```
ivansenic: wouldn't it be better to pass to complete config name to the superclass `exporters.metrics.prometheus.enabled`? Then you are not obligated to have property ending in `...enabled`.. Or how does this work actually?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/exporter/PrometheusExporterService.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Override
+    protected boolean checkEnabledForConfig(InspectitConfig conf) {
```
ivansenic: maybe `isEnabled` or ``isServiceEnabled` sounds better?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/exporter/PrometheusExporterService.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Override
+    protected boolean doEnable(InspectitConfig configuration) {
+        val config = configuration.getExporters().getMetrics().getPrometheus();
+        try {
+            String host = config.getHost();
+            int port = config.getPort();
+            log.info("Starting Prometheus Exporter on {}:{}", host, port);
+            PrometheusStatsCollector.createAndRegister(PrometheusStatsConfiguration.builder().setRegistry(defaultRegistry).build());
+            prometheusClient = new HTTPServer(/*host*/ host, /*port*/  port, /*daemon*/ true);
```
ivansenic: is there a need for `/*host*/` and port, daemon comments?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/exporter/PrometheusExporterService.java
```diff
@@ -0,0 +1,59 @@
...
+        } catch (Exception e) {
+            log.error("Error Starting Prometheus HTTP Endpoint!", e);
+            defaultRegistry.clear();
```
ivansenic: would this clear any other provider that could be registered? `io.prometheus.client.CollectorRegistry.defaultRegistry` seems global stuff? So can it be that we can not have at same time two exporters? because I know OC supports this and so should we..

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/exporter/PrometheusExporterService.java
```diff
@@ -0,0 +1,59 @@
...
+
+import static io.prometheus.client.CollectorRegistry.defaultRegistry;
+
+/**
+ * Service for the Prometheus OpenCensus exporter.
+ * Can be dynamically started and stopped using the exporters.metrics.prometheus.enabled configuration.
+ */
+@Component
+@Slf4j
+public class PrometheusExporterService extends DynamicallyActivatableService {
```
ivansenic: I think we should have a Spring integration test for this class.. So something where you have first disabled, then you update the config, check it's enabled and then disable again.. Should not be too hard, but we would actually test multiple things.. What do you think?

inspectit-oce-core/src/main/resources/config/default.yml
```diff
@@ -17,6 +17,20 @@ inspectit:
       # if true the directory will be watched for changes. When changes occur, the configuration is automatically reloaded
       watch: true
 
+  #settings for configuring OpenCensus stats and trace exporters
```
ivansenic: do you think we need some kind of ultimate prop for metrics and tracing, for example:

`inspectit.metrics.enabled=true`

When this would be false nothing related to metrics would be done, like an electricity switch.. Then you don't need to configure set of other metric related things like no exporter, no this metric, no that metrics, etc..

inspectit-oce-core/src/test/java/rocks/inspectit/oce/core/SpringTestBase.java
```diff
@@ -0,0 +1,132 @@
...
+            ConfigurableEnvironment defaultEnv = ctx.getEnvironment();
+
+            testPropertySources = defaultEnv.getPropertySources().stream()
+                    .filter(ps -> ps.getName() != StandardEnvironment.SYSTEM_ENVIRONMENT_PROPERTY_SOURCE_NAME)
+                    .filter(ps -> ps.getName() != StandardEnvironment.SYSTEM_PROPERTIES_PROPERTY_SOURCE_NAME)
+                    .collect(Collectors.toList());
+
+            InspectitEnvironment env = new TestInspectitEnvironment(ctx);
+            ctx.setEnvironment(env);
+            //TODO: mock Instrumentation class and add to the context
```
ivansenic: Would this really be needed? maybe we should only target test with `Instrumentation` for the system tests.. Not sure.. but anyway you can keep the comment..

inspectit-oce-core/src/test/java/rocks/inspectit/oce/core/SpringTestBase.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.stream.Collectors;
+
+import static org.mockito.Mockito.*;
+
+/**
+ * Base class for all tests.
+ */
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration(classes = SpringConfiguration.class, initializers = SpringTestBase.TestContextInitializer.class)
+@TestPropertySource(properties = {
```
ivansenic: would it work if I would add such `TestPropertySource` to the classes that extend this one?

inspectit-oce-core/src/test/java/rocks/inspectit/oce/core/exporter/PrometheusExporterTest.java
```diff
@@ -0,0 +1,88 @@
...
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.springframework.test.annotation.DirtiesContext;
+import rocks.inspectit.oce.core.SpringTestBase;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class PrometheusExporterTest extends SpringTestBase {
```
ivansenic: let's call this `PrometheusExporterIntTest` and have a clear separation between unit and integration tests?

inspectit-oce-core/build.gradle
```diff
@@ -54,6 +55,12 @@ dependencies {
             "io.prometheus:simpleclient_httpserver:${prometheusClientVersion}"
     )
 
+    testImplementation(
+            'org.springframework:spring-test:5.1.3.RELEASE',
+            'org.apache.httpcomponents:httpclient:4.5.6',
+            'org.mockito:mockito-core:2.23.4'
```
ivansenic: can you add `assertj` as well, you will like it :)

http://joel-costigliola.github.io/assertj/

inspectit-oce-core/src/test/java/rocks/inspectit/oce/core/exporter/PrometheusExporterTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+    private static final int HTTP_TIMEOUT = 1000;
+
+    static CloseableHttpClient testClient;
```
ivansenic: Would we have any alternative here in comparison to the http client? Could we auto-wire the `PrometheusExporter` class and do some enabled/disabled check.. This is fine, but I ask because of other exporters that are actually having push approach so something like this can not be done..

inspectit-oce-core/src/test/java/rocks/inspectit/oce/core/SpringTestBase.java
JonasKunz: yes, it then overwrite the startup configuration.

inspectit-oce-core/src/test/java/rocks/inspectit/oce/core/exporter/PrometheusExporterTest.java
JonasKunz: You could also auto wire it but then you maybe won't notice that the endpoint is not working.
For other stuff where this is not applicable I think we can use Mockito @Spy on beans or do some even simpler checks.

PR #23, opened 2018-12-18
closes #5: added system tests to agent project

### LINKED ISSUE #5: Support for integration tests
Support for the integration test where a separate JVM is started with our agent and test is executed. A very good example is given in the https://github.com/census-instrumentation/opencensus-java/tree/master/contrib/agent, we could simply copy this.

Make sure:
- `opencensus-test` is on the test classpath, this provides utilities to test the OC exporters
- properties could be tampered with, -> we need a possibility to start integration test with f.e. `inspectit.oce.some-property=false` also if this is not default

Create a first test that serves as an example.

JonasKunz:
In addition also fixed a minor issue where the directory watch service was closed again even if it was already closed. This was not a critical problem but generated unnecessary log messages.


inspectit-oce-agent/build.gradle
```diff
@@ -21,4 +21,116 @@ jar {
...
+    .collect { new File(it).getCanonicalPath() }
+            .unique()
+
+    assert javaExecutables.size > 0 :
+            'No Java executables found for running system tests'
+
+    def agentJarPath = jar.archivePath.toString()
+
+    javaExecutables.eachWithIndex { javaExecutable, index ->
+        def perVersionSystemTest = task("systemTest_${index}", type: Test) {
```
ivansenic: nice stuff, can we maybe update the name of this new task to something like:

```
        def paths = "${javaExecutable}".split(File.separator)
        def javaName = paths.length > 2 ? paths[paths.length - 3] : "${index}"
        def perVersionSystemTest = task("systemTest_${javaName}", type: Test) {
```
this way I would have task as `systemTest_java-8-oracle` instead of  `systemTest_0`

ivansenic: does not work when you have two jre paths: `/usr/lib/jvm/java-8-openjdk-amd64/jre:/usr/lib/jvm/java-8-oracle/jre`, any idea how to solve this? if not let's keep it how it was :cry:

ivansenic: But I run this and it was all green:

`../gradlew  -PsystemTestJavaHomes="/usr/lib/jvm/java-11-openjdk-amd64:/usr/lib/jvm/java-8-openjdk-amd64:/usr/lib/jvm/java-8-oracle:/usr/lib/jvm/java-9-oracle" check`

inspectit-oce-agent/src/system-test/java/PrometheusMetricsExportSysTest.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.http.util.EntityUtils;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+import java.util.Collections;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class PrometheusMetricsExportSysTest {
```
ivansenic: Hmm, I am asking my self if this would be the way to go... OK so first we already tested the Prometheus exporter in the IntTest..

I would find it much better that we are not testing the specific exporter here, but using something generic like `opencensus-test` lib.. Because we can only test whatever exporter is there per default config and as they can be switched to one that are not testing.

I checked the lib and there is a support for the span exporting testing with https://github.com/census-instrumentation/opencensus-java/blob/master/testing/src/main/java/io/opencensus/testing/export/TestHandler.java.. There is no such thing for merics expoter, but I think this is because you don't need it.. You can simple get `Metrics.getExportComponent().getMetricProducerManager()` and iterate over all metrics with:

```
for (MetricProducer metricProducer : metricProducerManager.getAllMetricProducer()) {
        for (Metric metric : metricProducer.getMetrics()) {
```

What do you think, should we try this out.. The you can also easily assert of the properties (name, desc, unit) of your metric is correctly set and not only the responseText assertsion.

PR #28, opened 2018-12-24
 closes #12: logback logging configuration support

### LINKED ISSUE #12: Logging strategy and configuration
Let's copy the old logging strategy we had with log-back originally in inspectIT.

But:

* keep the log-back config in the classpath resources so it's used by default if the external file is not found
* include the log-back configuration to the released artifact next to the inspectit-agent.jar (as now, but this time only use this as a reference for the user)
* ensure that property exists that can point to some other location configuration file
* remove the `MinLog` that we used originally because of Kryo 

Maybe we should provide some easy logging config in the configuration file. This way a user can easily configure logging as everything else.

ivansenic:
@JonasKunz I am missing the automatic logging updates on properties change, I was not sure how can I have this expression annotation thingy set.. I would do this in the additional commit and add this feature in the documentation as well -> as soon as you show me the way :smile: .

Some info:
* I kept using the piccolo parser as in the original project, this way we are not interacting with system Sax parser and having everything we need in our own class loader..
* disabling of the console or file appenders is done with logback filters, I am not aware of better solution..

### DIFF
--- a/inspectit-oce-core/build.gradle
+++ b/inspectit-oce-core/build.gradle
@@ -41,16 +41,23 @@ dependencies {
     )
 
     implementation(
+            // spring releated
             'org.springframework.boot:spring-boot:2.1.1.RELEASE',
             'org.yaml:snakeyaml:1.23',
+            'javax.annotation:javax.annotation-api:1.3.2', //Required for @PostConstruct and @PreDestroy to work in Java9+
 
-            //data validation
+            // data validation
             'org.hibernate.validator:hibernate-validator:6.0.13.Final',
             'org.apache.tomcat.embed:tomcat-embed-el:9.0.13',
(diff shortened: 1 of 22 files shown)


JonasKunz:
Regarding the dynamic configuration changes:
It would be easiest to pass the ApplicationContext to the LogbackInitializer.init method.
There you can then [register a listener](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/ConfigurableApplicationContext.html#addApplicationListener-org.springframework.context.ApplicationListener-). The listener should be specialized for InspectitConfigChangedEvents. In this listener you the ncompare the relevant proeprties of the old configuration with the new ones. If a change has occurred, reinit the logging.

PR #29, opened 2018-12-26
closes #8: support for common tags and tag providers

### LINKED ISSUE #8: Concept for the default tag context
How are we initializing the default tags, so generic application tags (service-name, host-name, process-id, etc). What tags we want to collect. The goal should be zero-configuration (no need to specify anything like before `-Dinspectit.agent.name`). We also need a way to collect tags that could come from a specific application type (SpringBoot) or cloud provider (CloudFundry, Azure, Kubernetes). 

In the scope of this ticket:
* create a generic tag provider interface / base class; these should also have a priority for overwriting (f.e. if the cloud foundry sets `VCAP_APPLICATION` env var, we should use service-name from this, not the properties or default)
* implement tag provider(s) that reads values from properties (name?)
* implement tag provider(s) that reads values from system (hostname, ip-address, etc)
* showcase how priority works, maybe by creating something that works for `SpringBoot`
* let the user have a possibility of adding a map of default tags via properties..

Kamon has a nice example on what can be defined in https://github.com/kamon-io/Kamon/blob/master/kamon-core/src/main/resources/reference.conf

ivansenic:
Example usages:
* creating view
 ```
        View.create(View.Name.create("some/name"), "some desc", someMeasure, someAggregation, tagsProvider.getCommonTagKeys())
 ```
* recording metrics view
 ```
        try (Scope scope = tagsProvider.withCommonTagScope()) {
            Stats.getStatsRecorder().newMeasureMap().put(someMeasure, 1L).record();
        }
 ```

@JonasKunz let's disucss how should we document this

**Main challendge: How can common tags be included in metrics collected by the user using instrumentation lib?**
As OpenCensus does not provide any mechanism to set common tags, we would need to allow user to access them.. Maybe we can provide some bootstrap interface for this, that user could use to get it, but this way the bootstrap must be a compile depdency for  users as well. We would need to provide NoOp impl as well. I had idea to instrument all threads and add  `try (Scope scope = tagsProvider.withCommonTagScope())` as the first statemetnt on the `run()`, but this would work only for recording, not for view registration.

### DIFF
--- a/inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/tags/CommonTagsManager.java
+++ b/inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/tags/CommonTagsManager.java
@@ -1,0 +1,108 @@
+package rocks.inspectit.oce.core.tags;
+
+import io.opencensus.common.Scope;
+import io.opencensus.tags.*;
+import org.springframework.stereotype.Component;
+
+import java.util.*;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+/**
+ * Component that provides tags that should be considered as common and used when ever a metric is recorded.
+ */
+@Component
+public class CommonTagsManager {
+
+    /**
+     * All {@link ITagsProvider}s registered in the manager.
+     */
+    private final List<ITagsProvider> providers = new CopyOnWriteArrayList<>();
+
+    /**
+     * OpenCensus tag context representing common tag context.
+     */
+    private TagContext commonTagContext;
+
+    /**
+     * List of common tag keys that can be used when creating the views.
+     */
+    private List<TagKey> commonTagKeys;
+
+    /**
+     * Returns common tags keys that all view should register.
+     *
+     * @return Returns common tags keys that all view should register.
+     */
+    public List<TagKey> getCommonTagKeys() {
+        return Collections.unmodifiableList(commonTagKeys);
+    }
+
+    /**
+     * Returns newly created scope with common tag context. Metrics collectors should use this Scope with the try/resource block:
+     * <code>
+     * try (Scope scope = withCommonTagScope()) {
+     * Stats.getStatsRecorder().newMeasureMap().put(M_ERRORS, 1L).record();
+     * }
+     * </code>
+     *
+     * @return Returns newly created scope with default tag context.
+     */
+    public Scope withCommonTagScope() {
+        return Tags.getTagger().withTagContext(commonTagContext);
+    }
+
+    /**
+     * Returns default tag context. For testing.
+     *
+     * @return Returns default tag context.
+     */
+    TagContext getCommonTagContext() {
+        return commonTagContext;
+    }
+
+    /**
+     * Registers a tag provider.
+     *
+     * @param tagsProvider ITagProvider
+     */
+    public void register(ITagsProvider tagsProvider) {
+        if (!providers.contains(tagsProvider)) {
+            providers.add(tagsProvider);
+            createDefaults();
+        }
+    }
+
+    /**
+     * Unregisters a tag provider.
+     *
+     * @param tagsProvider ITagProvider
+     */
+    public void unregister(ITagsProvider tagsProvider) {
+        if (providers.contains(tagsProvider)) {
+            providers.remove(tagsProvider);
+            createDefaults();
+        }
+    }
+
+    /**
+     * Processes all {@link #providers} on the start-up.
+     */
+    protected void createDefaults() {
+        // first create map of tags based on the providers priority
+        Map<String, String> all = new HashMap<>();
+        providers.stream()
+                .sorted(Comparator.comparingInt(ITagsProvider::getPriority).reversed())
+                .forEach(provider -> provider.getTags().forEach(all::putIfAbsent));
+
+        // then create key/value tags pairs for resolved map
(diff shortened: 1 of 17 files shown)


inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/tags/CommonTagsManager.java
```diff
@@ -0,0 +1,108 @@
...
+        if (providers.contains(tagsProvider)) {
+            providers.remove(tagsProvider);
+            createDefaults();
+        }
+    }
+
+    /**
+     * Processes all {@link #providers} on the start-up.
+     */
+    protected void createDefaults() {
```
JonasKunz: Adjust the comment and maybe rename it? like "collectTags" maybe

inspectit-oce-core/src/main/resources/config/default.yml
```diff
@@ -3,6 +3,23 @@ inspectit:
...
+        # if environment provider is enabled
+        enabled: true
+        # should the host name be resolved using InetAddress.getLocalHost(), if false 'host' tag is not added by env provider
+        resolve-host-name: true
+        # should the host address be resolved using InetAddress.getLocalHost(), if false 'host-address' tag is not added by env provider
+        resolve-host-address: true
+
+    # specifies user defined tag keys and values as a map
+    # these tag values would overwrite any value added by the providers, thus you can easily overwrite tags values by your own
+    # extra:
```
JonasKunz: have you maybe tried "extra: {}"?

inspectit-oce-documentation/src/docs/asciidoc/_includes/monitoring-environment/common-tags.adoc
```diff
@@ -0,0 +1,69 @@
+=== Common Tags
+
+inspectIT OCE provides a map of common tag keys and values that are used when recording all metrics.
+This features enables to "label" metrics collected with inspectIT OCE and share common information about a process or an environment where the process runs.
```
JonasKunz: This feature (without s) .. and **to** share common information ...

inspectit-oce-documentation/src/docs/asciidoc/_includes/monitoring-environment/common-tags.adoc
```diff
@@ -0,0 +1,69 @@
...
+The map can be <<User Defined Tags Provider,extended or overwritten by the user>> when this is required.
```
JonasKunz: The link is broken

inspectit-oce-documentation/src/docs/asciidoc/_includes/monitoring-environment/common-tags.adoc
```diff
@@ -0,0 +1,69 @@
...
+
+Tags providers are responsible for extracting information from a specific source and provide a map of key value pairs that are then combined into the common tag context.
```
JonasKunz: Tags Providers or Tag Provider? Latter one sounds somehow a bit better, even though one provider can provide multiple tags...

inspectit-oce-documentation/src/docs/asciidoc/_includes/monitoring-environment/common-tags.adoc
```diff
@@ -0,0 +1,69 @@
...
+Each provider specifies a priority and if same tag keys are provided by two providers, then the tag value from a provider with higher priority will be used.
```
JonasKunz: .. and if **the** same tag keys are provided by two providers, then the tag value from **the** provider with higher priority will be used.

inspectit-oce-core/src/main/resources/config/default.yml
ivansenic: haha, you got it.. nice :beers:

inspectit-oce-documentation/src/docs/asciidoc/_includes/monitoring-environment/common-tags.adoc
```diff
@@ -0,0 +1,69 @@
...
+----
+inspectit:
+  tags:
+    extra:
+      region: us-west-1
+      stage: preprod
+----
+
+==== Environment Tags
+The environment tags provider extends the common tag context by supplying information about the service name and a host.
```
JonasKunz: .. and the (network?) host

inspectit-oce-documentation/src/docs/asciidoc/_includes/monitoring-environment/common-tags.adoc
```diff
@@ -0,0 +1,69 @@
...
+The service name is resolved from the `inspectit.service-name` property, while host information is extracted using `InetAddress.getLocalHost()`.
```
JonasKunz: , while **the** host information

inspectit-oce-documentation/src/docs/asciidoc/_includes/monitoring-environment/common-tags.adoc
```diff
@@ -0,0 +1,69 @@
...
+
+NOTE: Note that on machines that have multiple network interfaces the `InetAddress.getLocalHost()` method might not provide desired values for host related tags.
```
JonasKunz: NOTE: Note that is double. Just write:
"NOTE: On machines that ..."

PR #33, opened 2018-12-28
closes #13: self-monitoring concept

### LINKED ISSUE #13: Self monitoring concept
Add ability to include our own metrics and export them in the same way as the application metrics are exported. Have an own namespace for this, f.e. `inspectit.internal`.

* can be turned off/on completely, by default disabled
* integrate into metrics collector(s) and expose collection operation time for example

OC also provides some kind of self-monitoring concept. Maybe we should check how we could be integrated there.

ivansenic:


inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/selfmonitoring/SelfMonitoringSettings.java
```diff
@@ -0,0 +1,23 @@
...
+@NoArgsConstructor
+public class SelfMonitoringSettings {
+
+    /**
+     * If self-monitoring is enabled.
+     */
+    private boolean enabled;
+
+    /**
+     * What should be created inspectIT self-monitoring measure.
```
JonasKunz: This comment is not understandable to me

PR #37, opened 2018-12-30
Closes #11: Created docker-based demo using Spring Pet Clinic

### LINKED ISSUE #11: Docker compose for fast start-up
Immediately create a docker compose that can be used to start everything with our agent. This should serve as the quick start for any user that wants to try inspectit-oce in under 3 minutes. Check if we need another docker compose for the development environment (I guess not as zipkin docker image would be enough) .

Components:
- demo-app with agent (this should be optional -> maybe user want's it's own app)
- prometheus 
- zipkin
- grafana

As we need to showcase that exporting to influxDB is possible, another setup could be:
- demo-app with agent (this should be optional -> maybe user want's it's own app)
- telegraph
- influxdb 
- zipkin
- grafana

Make sure all components are correctly connected.

AlexanderWert:
This pull request depends on a release of the inspectit OCE agent being available under the following URL:
https://github.com/inspectIT/inspectit-oce/releases/download/0.1.M1/inspectit-oce-agent-0.1.M1.jar

So, please merge this pull requests earliest, when the first release (version 0.1.M1) has been created.

Cheers,
Alex


inspectit-oce-documentation/build.gradle
```diff
@@ -17,7 +17,8 @@ asciidoctor {
     attributes 'inspectit-oce-version': project.version,
             'opencensus-version': "${openCensusVersion}",
             'gradle-rootdir-absolute': rootDir.absolutePath,
-            revnumber: project.version
+            revnumber: project.version,
+            toc: 'left'
```
ivansenic: naaahh I did no like the left TOC due to the length of the headings, but fine..

inspectit-oce-documentation/src/docs/asciidoc/_includes/getting-started/docker-compose-example.adoc
```diff
@@ -1,6 +1,121 @@
...
+We use the distributed version of the https://github.com/spring-petclinic/spring-petclinic-microservices[Spring PetClinic sample application] as the target application in our demo.
+To demonstrate the flexibility of the OpenCensus bases inspectIT agent, we provide different demo scenarios covering different monitoring and APM components:
 
-`// TODO Alex write about this docker example stuff..`
\ No newline at end of file
+
+
+.Demo Scenarios
+|===
+|docker-compose file | Landscape | Description
```
ivansenic: **D**ocker-compose?

inspectit-oce-demo/docker-compose-elastic.yml
```diff
@@ -0,0 +1,160 @@
...
+    image: openapm/spring-petclinic-config-server:inspectit-oce-demo
+    container_name: config-server
+    environment:
+     - INSPECTIT_SERVICE_NAME=config-server
+     - INSPECTIT_EXPORTERS_METRICS_PROMETHEUS_ENABLED=true
+     - INSPECTIT_EXPORTERS_METRICS_PROMETHEUS_HOST=0.0.0.0
+     - INSPECTIT_EXPORTERS_METRICS_PROMETHEUS_PORT=9096
+    mem_limit: 512M
+    entrypoint: ["sh", "-c","
+      wget --no-check-certificate https://github.com/inspectIT/inspectit-oce/releases/download/${INSPECTIT_OCE_VERSION}/inspectit-oce-agent-${INSPECTIT_OCE_VERSION}.jar -O inspectit-oce-agent.jar
```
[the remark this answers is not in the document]
JonasKunz: We agreed on the following: The fetching of the release is currently fine, as it allows the shipping of a minimal demo project. In the future, we however will include a separate docker-compose script which uses the locally built agent.

inspectit-oce-documentation/build.gradle
JonasKunz: I agree with Alex' version here, makes it much easier to navigate

JonasKunz:
I have adressed the fixes and the adding of volumes for file based configuration in two different commits. So if something does not work, we can always use only fixes without the volumes addition.

inspectit-oce-demo/docker-compose-elastic.yml
```diff
@@ -0,0 +1,166 @@
...
+      -jar /app.jar"]
+    volumes:
+     - ./agentconfig/config-server:/agentconfig
+    ports:
+     - 8888:8888
+     - 9096:9096
+
+  discovery-server:
+    image: openapm/spring-petclinic-discovery-server:inspectit-oce-demo
+    container_name: discovery-server
```
ivansenic: all hostnames are defined by docker and are given unique UUID, this creates different hosts in system dashboard although it's actually the same host and it should be only one available as it runs on the same machine. Luckily we can solve this easily by adding the `hostname: whatever` property to the services that are monitored by the agent (so excluding the influx, elastic, etc). If we use the same hostname in all services, we end up with just one host in the dashboard. Should we make this change?

I tried this and it works:

```
  api-gateway:
    image: openapm/spring-petclinic-api-gateway:inspectit-oce-demo
    container_name: api-gateway
    hostname: inspectit-oce-demo-host
```

inspectit-oce-demo/agentconfig/api-gateway/config.yml
```diff
@@ -0,0 +1,176 @@
+#inspectit:
```
ivansenic: hmm, is it okey to have a full config? I mean this represents then a very difficult point for maintaining.. Every time we change a default config file we would need to change this as well?

One solution could be to use symbolic links to the default config, but this is not working as well as the stuff would not be completely commented-out. 

What about if we just have a file with some `# comment` explaining what to do. I am also fine to create an empty hidden file (`.empty`) in these folders, so that at least folder structure is created correctly.

AlexanderWert: The config file is completely commented out! So it's like an empty file.
It's also how other tools are doing this, so the user at least sees all the configuration options.

PR #40, opened 2019-01-04
closes #18: Added handling of invalid startup configurations

### LINKED ISSUE #18: Add configuration validation
Validate the configuration on changes in the `ConfigurationCenter`. We should use utilities from Spring that should make this quite simple. Don't update nor fire the update event if the new configuration is not valid.

JonasKunz:
The configuration loading process has been altered as follows.
Intially, the agent only loads the "inspectit.config" part from the configuration for parsing all configuration sources.
Afterwards, it tries to load the entrie configuration. If this fails, it will use a fallback configuration where everything is disabled. This fallback configuration however will retain the previously loaded "inspectit.config" part, which means that config updates can be passed to the agent.

### DIFF
--- a/inspectit-oce-core/src/main/resources/config/fallback.yml
+++ b/inspectit-oce-core/src/main/resources/config/fallback.yml
@@ -1,0 +1,8 @@
+#------------------------
+# This file contains fallback configuration overwrites in case the agent is started with an invalid configuration
+# In this case, the agent tries to preserve the inspecit.config settings and to attach them to the fallback configuration
+# to make sure that runtime configuration updated are possible without a restart
+#------------------------
+inspectit:
+  metrics:
+    enabled: false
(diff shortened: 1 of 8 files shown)


inspectit-oce-core/src/main/resources/config/fallback.yml
```diff
@@ -0,0 +1,8 @@
+#------------------------
```
ivansenic: what was the reasoning for having the fallback config that has nothing active? i mean the other option I see is to have the default one on. But I guess I am overlooking something..

inspectit-oce-core/src/test/java/rocks/inspectit/oce/core/config/FallbackConfigTest.java
```diff
@@ -0,0 +1,23 @@
...
+@TestPropertySource(properties = {
+        "inspectit.thread-pool-size=iAmNotANumber"
+})
+public class FallbackConfigTest extends SpringTestBase {
+
+    @Autowired
+    InspectitEnvironment env;
+
+    @Test
+    public void testFallbackConfigurationActive() {
```
ivansenic: can you add a test that makes sure that when an invalid update occurs we remain with the previous config and not the fallback?

PR #41, opened 2019-01-07
Improved CI configuration - Testing on multiple JDKs in parallel and deploying documentation to S3

mariusoe:
See the following build as a sample for this configuration: https://travis-ci.com/mariusoe/inspectit-oce/builds/96524934

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,97 @@
 language: java
-jdk: openjdk8
 dist: trusty
+jdk:
+  - oraclejdk8
+  - oraclejdk9
+  - oraclejdk11
+  - openjdk8
 
 #enable caching
 before_cache:
   - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
   - rm -fr $HOME/.gradle/caches/*/plugin-resolution/
+
 cache:
   directories:
     - $HOME/.gradle/caches/
     - $HOME/.gradle/wrapper/
-    
+    - $HOME/ibm-jdk-installer/
+
+install: true
 script:
-  - wget https://raw.githubusercontent.com/sormuras/bach/76983e2f99497738b6f3e9f86bd42fd3a759a275/install-jdk.sh -O $TRAVIS_BUILD_DIR/install-jdk.sh
-  - bash $TRAVIS_BUILD_DIR/install-jdk.sh --feature 11 -L GPL --target $HOME/openjdk11
-  - ./gradlew build codeCoverageReport -PsystemTestJavaHomes="$JAVA_HOME:$HOME/openjdk11"
-after_success:
-  # upload coverage report
-  - bash <(curl -s https://codecov.io/bash)
+  - ./gradlew :inspectit-oce-agent:systemTest
+
+jobs:
+  include:
+    - name: "JDK ibmjdk8 Java"
+      sudo: true
+      install:
+        - wget -nc -O $HOME/ibm-jdk-installer/ibm-sdk.bin https://s3-eu-west-1.amazonaws.com/inspectit-oce-files/ibm-java-sdk-8.0-5.26-x86_64-archive.bin || true
+        - sudo chmod +x $HOME/ibm-jdk-installer/ibm-sdk.bin
+        - echo "INSTALLER_UI=silent" > $HOME/ibm-install.properties
+        - echo "USER_INSTALL_DIR=$HOME/ibm-jdk-8" >> $HOME/ibm-install.properties
+        - echo "LICENSE_ACCEPTED=TRUE" >> $HOME/ibm-install.properties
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -10,11 +13,33 @@ cache:
...
+jobs:
+  include:
+    - stage: deploy
+      name: "Code Coverage Reporting"
+      jdk: oraclejdk8
+      install: ./gradlew build codeCoverageReport
+      after_success:
+        # upload coverage report
+        - bash <(curl -s https://codecov.io/bash)
+    - name: "Documentation Deployment"
```
JonasKunz: Don't we also need the "stage" here? Or does is copy the stage from the previous job?

.travis.yml
```diff
@@ -10,11 +13,33 @@ cache:
...
+      jdk: oraclejdk8
+      install: ./gradlew :inspectit-oce-documentation:asciidoctor
+      deploy:
+        provider: s3
```
JonasKunz: Would it be possible to upload the documentation with e.g. a build number in its path, so that the documentation for every PR is accessible? We would then of course most likely need some expiration so that we don't pollute the S3.

What we definetely need is two paths: one for the master and one for the latest release.

.travis.yml
```diff
@@ -1,6 +1,9 @@
 language: java
-jdk: openjdk8
 dist: trusty
+jdk:
```
ivansenic: I would really like to have an IBM JVM here.. I am quite sure this is not so easily possible as just listing it here, but maybe some approach as Jonas had before with openjdk11? testing different versions is fine, but we also need to split on the providers (IBM, openjdk, etc).. We can have this in separate ticket, but this is important to have as early as possible as then we recognize problems right away..

.travis.yml
```diff
@@ -10,11 +13,33 @@ cache:
...
+      jdk: oraclejdk8
+      install: ./gradlew :inspectit-oce-documentation:asciidoctor
+      deploy:
+        provider: s3
+        access_key_id: $AWS_ACCESS_KEY_ID
+        secret_access_key: $AWS_SECRET_ACCESS_KEY
+        bucket: "docs.inspectit.rocks"
+        region: eu-west-1
+        skip_cleanup: true
+        local_dir: inspectit-oce-documentation/build/asciidoc/html5
```
ivansenic: if you are not using the zipped version that is released, maybe you can remove this part from the build file of the documentation as zipping has no meaning.. I added this because of the upload, but was not aware we would be able to simply point to the complete dir..

ivansenic: Actually, I just realized that with this PR we would have only one documentation available and we would like to have versioned documentation. What does this mean:

1. I want to have documentation for each released version, this can be achieved by having the folder with the version name `.../0.1.M1` (similar to Spring documentation)
2. In addition, we should have a `.../current` folder that shows the documentation for the latest release. This would be used as the link to the official documentation and hopefully index by google as the first choice.

This is why in the ZIP task I added a parent folder with the version name so that unpacking would solve the point 1 above.

.travis.yml
mariusoe: Yes, that's just the name. The stage will be the previous one

.travis.yml
mariusoe: Done.

.travis.yml
mariusoe: Let's don't do this yet. Sure you have to build the docs by yourself but otherwise we do have to do all this clean-up in the end.

.travis.yml
mariusoe: Removed zip task

mariusoe: And a documentation will be kept for each released version

ivansenic: OK I kind of changed my mind. Let's keep the zip task and upload the documentation zip also as the release artifact... I really like how it is now in the first release, that Jonas uploaded the agent jar and documentation zip.. That's cool imo, that we also provide this irrelevant of our docs website..

.travis.yml
```diff
@@ -1,20 +1,97 @@
...
-  - bash <(curl -s https://codecov.io/bash)
\ No newline at end of file
+  - ./gradlew :inspectit-oce-agent:systemTest
+
+jobs:
+  include:
+    - name: "JDK ibmjdk8 Java"
+      sudo: true
+      install:
+        - wget -nc -O $HOME/ibm-jdk-installer/ibm-sdk.bin https://s3-eu-west-1.amazonaws.com/inspectit-oce-files/ibm-java-sdk-8.0-5.26-x86_64-archive.bin || true
```
ivansenic: the build is failing again, I guess this is because of the GC test, can we please merge #43 first then and then prove that this is green?

mariusoe: Yes, that's the reason

PR #43, opened 2019-01-09
Closed #42 - improving system test for GC metric recorder

### LINKED ISSUE #42: Improving system test for GCMetricsRecorder
The system test for the `GCMetricsRecorder` is often failing due to a missing GC. The test is depending on a GC in a certain timeframe. This GC will be enforce but it is not always working resulting in failing test - even the actual code is properly working.

The test should be more robust and be able to handle if no GC is happening in a certain duration without leting the build failing

mariusoe:


inspectit-oce-agent/src/system-test/java/rocks/inspectit/oce/metrics/GCMetricsSysTest.java
```diff
@@ -38,6 +40,11 @@ public void testGCPauseCapturing() throws Exception {
             System.gc();
             System.runFinalization();
             if (objRef.get() != null) {
+                // even more trash
```
ivansenic: what's we the weak reference idea?

ivansenic: should we also decrease the default metrics collection frequency for the system tests (I am not sure if this has been done, but could this also be a problem here)... Default is 15s?

JonasKunz: The GC metrics recorder is not polling based, therefore the frequency does not impact it. The recorder performs its duty based on events sent by the JVM after a GC.

ivansenic: fancy

inspectit-oce-agent/src/system-test/java/rocks/inspectit/oce/metrics/GCMetricsSysTest.java
```diff
@@ -19,6 +19,8 @@
 
     private static final ViewManager viewManager = Stats.getViewManager();
 
+    public static List blackhole;
```
JonasKunz: What is the purpose of storing the trash in a variable?

mariusoe: When using a loop like this

    for (int i = 0; i < 1000000; i++) {
        new ArrayList<>();
    }
        
the compiler notices that the constructed object is not assigned and not used at all, thus, it will (most likly) "optimize" it away. So, during the test no garbage will be produced anymore.
Having a variable like this, the compile cannot know if this variable is and will be used because basically, it could/can be used by any class due to its `public` modifier - thus it cannot be safely removed by the optimizer.

Issue #54, opened 2019-01-14
Introduce the byte buddy to the core


ivansenic:
* byte buddy should be available to the Spring context
* we'll need a way to have a generic byte buddy configuration, check if there is anything we would need to configure at the moment

PR #76, opened 2019-01-16
Closes #56, #54 - Added ByteBuddy and instrumentation for Executors

### LINKED ISSUE #56: Special sensor for executor services
Target to instrument one method only, I think the old inspectIT has a correct configuration for this, please check, I think it's the interface based. GRPC context has a wrapping runnable utility.

* provide system test
* provide a config property that is `enabled` by default
* add the `Special Sensors` chapter in the asciidoc

JonasKunz:


inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/InstrumentationManager.java
```diff
@@ -0,0 +1,241 @@
...
+     */
+    private AgentBuilder setupIgnoredClasses(AgentBuilder baseAgent) {
+
+        ClassLoader inspectITClassLoader = getClass().getClassLoader();
+
+        return baseAgent.ignore((type, classloader, module, clazz, pd) -> {
+            if (classloader == inspectITClassLoader) {
+                return true;
+            } else if (classloader == null) {
+                return IGNORED_BOOTSTRAP_PACKAGES.stream()
```
ivansenic: This is not enough, we need to check if the class was loaded by the bootstrap class loader, otherwise, you are excluding all classes from given packages, even if they are loaded by the application class loader.. That seems not correct imo.. Or

PR #78, opened 2019-01-18
Closes #56, #54 - Added asynchronous instrumentation capabilities and executor context propagation

### LINKED ISSUE #56: Special sensor for executor services

JonasKunz:
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-oce/78)


JonasKunz:
I pushed some refactorings.
Please have a look at the code and feel free to propose further changes, especially regarding the names of the stuff as I am terrible at naming abstract thingys.
Test are not implemented yet, they will be added in this PR as soon as the overall structure of the implementation is stable
I would recommend that we add the self-monitoring in a separate PR, as it requires several changes to the self-monitoring component.

JonasKunz:
Something i forgot to mention: when you try out the agent and play around with the configuration it is best to set the log level to debug. There you get detailed messages of the work which is being done.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/AsyncClassTransformer.java
```diff
@@ -0,0 +1,221 @@
...
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.BiConsumer;
+
+/**
+ * A class transformer applying all inspectIT instrumentations.
+ * This transform only instrument classes when they are redefined / retransformed!.
+ */
+@Component
+@Slf4j
+public class AsyncClassTransformer implements ClassFileTransformer {
```
ivansenic: we need a detailed unit test with mocked `Instrumentation`

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/AsyncClassTransformer.java
```diff
@@ -0,0 +1,221 @@
...
+            instrumentation.removeTransformer(this);
+        }
+    }
+
+
+    private byte[] applyInstrumentation(Class<?> classBeingRedefined, byte[] bytecode) {
+        try {
+            TypeDescription type = TypeDescription.ForLoadedType.of(classBeingRedefined);
+
+            ClassInstrumentationConfiguration classConf;
```
ivansenic: combined to one line?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/ClassInstrumentationConfiguration.java
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * The configuration representing that no instrumentation of the class is performed.
+     */
+    static final ClassInstrumentationConfiguration NO_INSTRUMENTATION = new ClassInstrumentationConfiguration(Collections.emptySet(), null);
+
+    /**
+     * The sensors which are active for the target class.
+     */
+    Set<SpecialSensor> activeSensors;
```
ivansenic: I would already extract this to some kind of the interface so that we immediately operate on the interface that `SpecialSensor`s are implementing.. This way we are ready for all the needed changes when we add a new sensor type..

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/ClassInstrumentationConfiguration.java
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * The inspectIT configuration used to derive this configuration object.
+     * Can be null, but only if no sensors active.
+     */
+    InstrumentationSettings activeConfiguration;
```
ivansenic: let's make all the fields `final` and `private`?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/ClassInstrumentationConfiguration.java
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * Compares this instrumentation configuration against another.
+     * Two instrumentations are considered to be the "same" if they result in the same bytecode changes.
+     *
+     * @param forType the type for which this configuration is meant.
+     * @param other   the other configuration to compare against
+     * @return true if both are the same, false otherwise
+     */
+    public boolean isSameAs(TypeDescription forType, ClassInstrumentationConfiguration other) {
+        if (!activeSensors.equals(other.activeSensors)) {
```
ivansenic: Here it's better to go for `Collections.disjoint` where you can find out if elements of in the collections are same or not.. this way you don't depend on the collection implementation to be same...

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/ClassInstrumentationConfiguration.java
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Builds the instrumentation configuration for a class based on the overall {@link rocks.inspectit.oce.core.config.model.InspectitConfig}.
+     *
+     * @param forType    the type for which the configuration shall be derived
+     * @param config     the active inspectIT config
+     * @param allSensors the list of all available sensors
+     * @return
+     */
+    public static ClassInstrumentationConfiguration getFor(TypeDescription forType, InstrumentationSettings config, Collection<? extends SpecialSensor> allSensors) {
```
ivansenic: should we have a private constructor in the class to enforce using this static init?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/NewClassDiscoveryService.java
```diff
@@ -0,0 +1,79 @@
...
+                classDiscoveryListeners.forEach(lis -> lis.accept(newClasses));
+            } else {
+                log.debug("No new classes found  check took {} ms (last class define {} ms ago)", watch.getElapsedMillis(), timeSinceLastClassDefinition);
+            }
+        }
+        scheduleUpdateCheck();
+    };
+
+
+    public void addDiscoveryListener(Consumer<Set<Class<?>>> listener) {
```
ivansenic: same as for the `AsyncListener` I would create the interface and use auto-wire for the listeners.. The benefit would be that you could already pass the known classes on the `init()`, this way the other component does not have to do it as well and should only listen to the updates..

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/NewClassDiscoveryService.java
```diff
@@ -0,0 +1,79 @@
...
+        classDiscoveryListeners.add(listener);
+    }
+
+    @PostConstruct
+    private void init() {
+        knownClasses.addAll(Arrays.<Class<?>>asList(instrumentation.getAllLoadedClasses()));
+        scheduleUpdateCheck();
```
ivansenic: do we need a `@Destory` to cancel the update check?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/special/ExecutorContextPropagationSensor.java
```diff
@@ -0,0 +1,55 @@
...
+
+@Component
+@DependsOn(ContextManagerImpl.BEAN_NAME)
+@Slf4j
+public class ExecutorContextPropagationSensor implements SpecialSensor {
+    @Override
+    public boolean shouldInstrument(TypeDescription type, InstrumentationSettings settings) {
+        val matcher = isSubTypeOf(Executor.class)
+                .and(not(isAbstract()))
+                .and(not(nameStartsWith("io.grpc.Context$")
```
ivansenic: shouldn't this already be ignored by or ignore rules?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/special/ExecutorContextPropagationSensor.java
```diff
@@ -0,0 +1,55 @@
...
+
+import static net.bytebuddy.matcher.ElementMatchers.*;
+
+@Component
+@DependsOn(ContextManagerImpl.BEAN_NAME)
+@Slf4j
+public class ExecutorContextPropagationSensor implements SpecialSensor {
+    @Override
+    public boolean shouldInstrument(TypeDescription type, InstrumentationSettings settings) {
+        val matcher = isSubTypeOf(Executor.class)
```
ivansenic: can we extract the byte buddy matcher to the `static` field, this seems to be constant and we are always recreating it?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/NewClassDiscoveryService.java
```diff
@@ -0,0 +1,79 @@
...
+
+@Service
+@Slf4j
+public class NewClassDiscoveryService {
+
+    @Autowired
+    private InspectitEnvironment env;
+
+    @Autowired
+    private AsyncClassTransformer transformer;
```
ivansenic: let's discuss how can we remove this dependency and decouple the components to be

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/service/BatchJobExecutorService.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.function.IntConsumer;
+
+/**
+ * Service to schedule asynchronous tasks which are executed in batches.
+ */
+@Service
+@Slf4j
+public class BatchJobExecutorService {
```
ivansenic: unit test pls..

inspectit-oce-core/src/main/resources/config/default.yml
```diff
@@ -34,6 +34,31 @@ inspectit:
...
+      executor-context-propagation: true
+    # settings for fine-tuning the instrumentation process
+    internal:
+      # defines how many classes are checked at once for updates of their configuration
+      class-configuration-check-batch-size: 1000
+      # the time to pause between executing batches of class instrumentation configuration checks
+      class-configuration-check-inter-batch-delay: 50ms
+
+      # defines how many classes are retransformed at once
+      class-retransform-batch-size: 100
```
ivansenic: this seems a lot..

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/NewClassDiscoveryService.java
```diff
@@ -0,0 +1,79 @@
...
+import java.time.Duration;
+import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
+
+@Service
+@Slf4j
+public class NewClassDiscoveryService {
```
ivansenic: unit test as well.. make sure to simulate new class loading in between new checks..

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/ClassRetransformationService.java
```diff
@@ -0,0 +1,145 @@
...
+
+    @Autowired
+    private AsyncClassTransformer transformer;
+
+    /**
+     * The queue of classes to process.
+     * For each element in the queue, this service checks if a retransformation is still required.
+     * If this is the case, then a retransformation is triggered.
+     */
+    private Set<Class<?>> pendingClasses = Collections.synchronizedSet(Collections.newSetFromMap(new WeakHashMap<>()));
```
ivansenic: guava cache?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/ClassRetransformationService.java
```diff
@@ -0,0 +1,145 @@
...
+public class ClassRetransformationService {
+
+    @Autowired
+    private BatchJobExecutorService executor;
+
+    @Autowired
+    private InspectitEnvironment env;
+
+    @Autowired
+    private List<SpecialSensor> specialSensors;
```
ivansenic: use interface as well..

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/ClassRetransformationService.java
```diff
@@ -0,0 +1,145 @@
...
+    public void requestRetransformation(Class<?> clazz) {
+        pendingClasses.add(clazz);
+    }
+
+    @PostConstruct
+    private void init() {
+        val settings = env.getCurrentConfig().getInstrumentation().getInternal();
+        Duration delay = settings.getClassConfigurationCheckInterBatchDelay();
+        int batchSize = settings.getClassRetransformBatchSize();
+        retransformJob = executor.startJob(this::retransformClasses, batchSize, delay, delay);
```
ivansenic: do we need `@Destroy` to cancel the task?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/ClassRetransformationService.java
```diff
@@ -0,0 +1,145 @@
...
+     * The goal is to get and remove batchSize classes from the queueand to invoke {@link Instrumentation#retransformClasses(Class[])} on them.
+     * <p>
+     * The package access is for testing.
+     */
+    void retransformClasses(int batchSize) {
+        val watch = new StopWatch();
+        if (!transformer.isShuttingDown()) {
+            Set<Class<?>> classesToRetransform = new HashSet<>();
+            try {
+                Optional<Class<?>> nextClass = CommonUtils.pollElementFromSet(pendingClasses);
```
ivansenic: does it make sense to pull one by one, how about we pull at least the `bacthSize`?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/ClassRetransformationService.java
```diff
@@ -0,0 +1,145 @@
...
+     *
+     * @param clazz
+     * @return
+     */
+    private boolean doesClassRequireRetransformation(Class<?> clazz) {
+        val typeDescr = TypeDescription.ForLoadedType.of(clazz);
+        InstrumentationSettings instrSettings = env.getCurrentConfig().getInstrumentation();
+        val requestedConfig = ClassInstrumentationConfiguration.getFor(typeDescr, instrSettings, specialSensors);
+
+        val activeConfig = transformer.getActiveClassInstrumentationConfiguration(clazz);
```
ivansenic: maybe the active configuration could be stored in some special service for this purpose, where the transformer would report on the instrumentation and others can ask for such things? We are not gaining much, so let's think..

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/InstrumentationUpdateService.java
```diff
@@ -0,0 +1,176 @@
...
+public class InstrumentationUpdateService {
+
+    @Autowired
+    private BatchJobExecutorService executor;
+
+    @Autowired
+    private InspectitEnvironment env;
+
+    @Autowired
+    private List<SpecialSensor> specialSensors;
```
ivansenic: interface?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/InstrumentationUpdateService.java
```diff
@@ -0,0 +1,176 @@
...
+
+    @Autowired
+    private SelfMonitoringService selfMonitoring;
+
+    @Autowired
+    private AsyncClassTransformer transformer;
```
ivansenic: if this class implements `IPostInstrumentationListener` and `IClassesDiscoveryListener` then you don't need these direct dependences.. Also most of the stuff in the `init()` can be removed..

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/InstrumentationUpdateService.java
```diff
@@ -0,0 +1,176 @@
...
+
+    @Autowired
+    private NewClassDiscoveryService classDisoveryService;
+
+    @Autowired
+    private Instrumentation instrumentation;
```
ivansenic: this should also be removed and classes always retrieved by the Listener.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/InstrumentationUpdateService.java
```diff
@@ -0,0 +1,176 @@
...
+
+    @Autowired
+    private ClassRetransformationService retransformationService;
+
+
+    private Set<Class<?>> pendingClasses = Collections.synchronizedSet(Collections.newSetFromMap(new WeakHashMap<>()));
```
ivansenic: guava cache?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/InstrumentationUpdateService.java
```diff
@@ -0,0 +1,176 @@
...
+        // all newly loaded classes have to be checked if they need instrumentation
+        classDisoveryService.addDiscoveryListener(pendingClasses::addAll);
+        // it is possible that we think a class is up to date while it is actually being instrumented with an
+        // outdated instrumentation configuration .for this reason we recheck every class after it has been instrumented.
+        transformer.addPostInstrumentationListener((clazz, conf) -> pendingClasses.add(clazz));
+
+        //all already loaded classes need to be checked if they need instrumentation
+        requestConfigUpdateForAllLoadedClasses();
+
+        startClassCheckerJob();
```
ivansenic: destroy for cancel?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/InstrumentationUpdateService.java
```diff
@@ -0,0 +1,176 @@
...
+    void checkClassesForConfigurationUpdates(int numberOfClassesToProcess) {
+        if (transformer.isShuttingDown()) {
+            return; //nothing to do anymore
+        }
+
+        val watch = new StopWatch();
+        try {
+            int processedClassesCount;
+            for (processedClassesCount = 0; processedClassesCount < numberOfClassesToProcess; processedClassesCount++) {
+                Optional<Class<?>> nextClass = CommonUtils.pollElementFromSet(pendingClasses);
```
ivansenic: same here, let's poll at least `numberOfClassesToProcess`, maybe then you don't even need the for loop?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/InstrumentationUpdateService.java
```diff
@@ -0,0 +1,176 @@
...
+        }
+    }
+
+    /**
+     * Todo: where to actually put this method... ? (Code duplication)
+     *
+     * @param clazz
+     * @return
+     */
+    private boolean doesClassRequireRetransformation(Class<?> clazz) {
```
ivansenic: we are doing the same thing twice, first here, then in the retransform service.. Not really sure if this is the part where you optimize for this situation where you can quickly change instrumentation twice.. If this is the case, I would remove this optimization which would also make the `ClassRetransformationService` actually not needed as we can request the instrumentation directly from here.. It would decrease the complexity, speed things up and the only downside is covering the situation that happens very very rare.. Let's please discuss this..

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/InstrumentationUpdateService.java
```diff
@@ -0,0 +1,176 @@
...
+ * This class is responsible for making sure that for every class the instrumentation
+ * eventually is applied as specified in the active {@link rocks.inspectit.oce.core.config.model.InspectitConfig}
+ * <p>
+ * Whenever a classes instrumentation configuration might have changes, this class gets put into this services working queue.
+ * This service is then responsible for filtering out the classes whose instrumentaiton actually has
+ * changed and submits them to the {@link ClassRetransformationService}.
+ */
+@Service
+@Slf4j
+public class InstrumentationUpdateService {
```
ivansenic: I would like a uni test for this and an int test as well.. In the int test we can focus that interaction between services work, that configuration changes are applied, etc.. Maybe also the destruction of the context if possible..

build.gradle
```diff
@@ -62,11 +64,16 @@ subprojects {
             jarCheckOutput.mkdirs()
             configurations.runtimeClasspath.each { file ->
                 def name = file.name
-                javaexec {
-                    classpath configurations.buildTools
-                    main = 'com.mindprod.jarcheck.JarCheck'
-                    args = ["$file", "1.0", "1.8"]
-                    standardOutput = new File(jarCheckOutput, "$name-check.log").newOutputStream()
+                def isExcluded = excludes.stream().filter({ excl ->
```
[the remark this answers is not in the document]
JonasKunz: Yes bytebuddy, contains some Java 9 classes I think, which are however somehow only loaded when Java 9 is present.. therefore I excluded it here

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/AsyncClassTransformer.java
```diff
@@ -0,0 +1,221 @@
...
+
+    @Autowired
+    private List<SpecialSensor> sensors;
+
+    /**
+     * Detects if the instrumenter is in the process of shutting down.
+     * When it is shutting down, no new instrumentations are added anymore, instead all existing instrumentations are removed.
+     */
+    @Getter
+    private boolean shuttingDown = false;
```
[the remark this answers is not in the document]
JonasKunz: We agreed on keeping the monitor as it also makes the access to the set of instrumented classes atomic.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/ClassInstrumentationConfiguration.java
JonasKunz: Done by lombok @Value.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/ClassInstrumentationConfiguration.java
JonasKunz: As we create the set ourselves, equals is fine and faster.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/ClassRetransformationService.java
JonasKunz: Set does not offer this functionality, however with fuava sets we can reuse the iterator.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/ClassRetransformationService.java
JonasKunz: Moved it to the InstrumentationManager

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/InstrumentationUpdateService.java
JonasKunz: The direct dependency to the transformer is still required to detect if it is shutting down.
Otherwise we constantly issue classes for retransform even though they will never be instrumented..

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/NewClassDiscoveryService.java
```diff
@@ -0,0 +1,102 @@
...
+        }
+        if (!isShuttingDown) {
+            scheduleUpdateCheck();
+        }
+    };
+
+
+    @Override
+    public void newClassDefined(String className, ClassLoader loader) {
+        lastNewClassDefinitionTimestamp.set(System.currentTimeMillis());
```
ivansenic: use `lazySet` here

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/NewClassDiscoveryService.java
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * Future for the scheduled {@link #updateCheckTask} to make it cancelable.
+     */
+    private Future<?> updateCheckFuture;
+
+    /**
+     * Stores the timestamp for the last time a class was defined.
+     */
+    private AtomicLong lastNewClassDefinitionTimestamp = new AtomicLong();
```
mariusoe: I think you don't need an ``AtomicLong`` here. A ``private volatile long`` would be sufficient because you just direclty set/read the variable which are atomic operations by default (but you need the volatile on long).

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/InstrumentationManager.java
```diff
@@ -0,0 +1,209 @@
...
+public class InstrumentationManager implements IClassDiscoveryListener {
+
+    @Autowired
+    private BatchJobExecutorService executor;
+
+    @Autowired
+    private InspectitEnvironment env;
+
+    @Autowired
+    private NewClassDiscoveryService classDisoveryService;
```
mariusoe: Can be removed - not used

ivansenic: I am totally confused now, how is this needed here when this class implements the `IClassDiscoveryListener`, how is the context actually starting as this seems to be a cyclic dependency?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/InstrumentationManager.java
```diff
@@ -0,0 +1,209 @@
...
+    private Instrumentation instrumentation;
+
+    @Autowired
+    private InstrumentationConfigurationResolver configResolver;
+
+    /**
+     * For each class we remember the applied instrumentation.
+     * This allows us to check if a retransform is required.
+     */
+    private Map<Class<?>, ClassInstrumentationConfiguration> activeInstrumentations =
```
ivansenic: same, what about using Cache directly?
and for the below ones?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/service/BatchJobExecutorService.java
```diff
@@ -0,0 +1,130 @@
...
+     *
+     * @param <T>             the batch size type, e.g. an Integer
+     * @param task            the actual job to process
+     * @param batchSizes      the size of the batches
+     * @param startDelay      the time to wait until the execution of the first batch
+     * @param interBatchDelay the pause between the batches
+     * @return the started {@link BatchJob}
+     */
+    public <T> BatchJob<T> startJob(BatchProcessor<T> task, T batchSizes, Duration startDelay, Duration interBatchDelay) {
+        BatchJob job = new BatchJob(task, batchSizes, interBatchDelay);
```
ivansenic: you are missing the generics here?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationConfigurationChangedEvent.java
```diff
@@ -0,0 +1,33 @@
...
+ * The new {@link rocks.inspectit.oce.core.config.model.InspectitConfig} is used to derive the {@link InstrumentationConfiguration},
+ * for which then this event is triggered.
+ */
+
+public class InstrumentationConfigurationChangedEvent extends ApplicationEvent {
+
+    /**
+     * The configuration which has previously been active.
+     */
+    @Getter
```
mariusoe: ``@Getter`` would also work on a class level, then you don't have to add it to all members.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationConfigurationResolver.java
```diff
@@ -0,0 +1,118 @@
...
+
+import javax.annotation.PostConstruct;
+import java.lang.instrument.Instrumentation;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * This class is responsible for deriving the {@link InstrumentationConfiguration} from teh {@link InstrumentationSettings}.
```
mariusoe: typo: teh

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/InstrumentationManager.java
```diff
@@ -0,0 +1,209 @@
...
+     * The set of classes which might need instrumentation updates.
+     * This service works through this set in batches.
+     */
+    private Set<Class<?>> pendingClasses = Collections.newSetFromMap(
+            CacheBuilder.newBuilder().weakKeys().<Class<?>, Boolean>build().asMap());
+
+    private BatchJobExecutorService.BatchJob<BatchSize> classInstrumentationJob;
+
+    @Value
+    private static class BatchSize {
```
ivansenic: static inner class to bottom of this class

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationConfigurationResolver.java
```diff
@@ -0,0 +1,118 @@
...
+            return true;
+        }
+
+        if (clazz.getClassLoader() == INSPECTIT_CLASSLOADER) {
+            return true;
+        }
+        if (clazz.getClassLoader() == null) {
+            String name = clazz.getName();
+            boolean isIgnored = env.getCurrentConfig().getInstrumentation().getIgnoredBootstrapPackages().entrySet().stream()
+                    .filter(e -> e.getValue() == true)
```
mariusoe: can be simplified to ``filter(e -> e.getValue())`` or ``.filter(Map.Entry::getValue)``

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationConfigurationResolver.java
```diff
@@ -0,0 +1,118 @@
...
+                    .anyMatch(e -> name.startsWith(e.getKey() + "."));
```
mariusoe: Can we remove the attachment of the "." because then we could also exclude multiple package which starting with the same specified string.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/InstrumentationManager.java
```diff
@@ -0,0 +1,209 @@
...
+
+        InternalSettings newInternal = ev.getNewConfig().getInstrumentation().getInternal();
+        val batchSizes = new BatchSize(newInternal.getClassConfigurationCheckBatchSize(), newInternal.getClassRetransformBatchSize());
+        classInstrumentationJob.setBatchSizes(batchSizes);
+        classInstrumentationJob.setInterBatchDelay(newInternal.getInterBatchDelay());
+    }
+
+    @EventListener
+    private void instrumentationConfigEventListener(InstrumentationConfigurationChangedEvent ev) {
+        pendingClasses.addAll(loadedClasses);
```
ivansenic: Now we are holding twice the complete loaded classes in a set, here and in the discovery service.. would it make sense that we move this event listener to the discovery service and report via the listener? this way we would keep loaded classes only in the discovery service.. maybe we should then rename the `newClassesDiscovered(Set<Class<?>> newClasses)` to something more meaningful?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/InstrumentationManager.java
```diff
@@ -0,0 +1,209 @@
...
+    /**
+     * Processes a given amount of classes from {@link #pendingClasses}.
+     * For the classes where it is required a retransform is triggered.
+     * <p>
+     * This method is package-private for testing.
+     *
+     * @param batchSize the number of classes to take from {@link #pendingClasses} and to retransform per batch
+     */
+    void checkClassesForConfigurationUpdates(BatchSize batchSize) {
+        if (transformer.isShuttingDown()) {
```
ivansenic: the only reason we are auto-wiring the `transformer` is because of the shut down.. I see we also keep track of the shut-down in the `NewClassDiscoveryService`.. Can you quickly check if Spring offers some helper for context shutting down that we would use?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/ClassInstrumentationConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+    /**
+     * Compares this instrumentation configuration against another.
+     * Two instrumentations are considered to be the "same" if they result in the same bytecode changes.
+     *
+     * @param forType the type for which this configuration is meant.
+     * @param other   the other configuration to compare against
+     * @return true if both are the same, false otherwise
+     */
+    public boolean isSameAs(TypeDescription forType, ClassInstrumentationConfiguration other) {
+        if (!activeSpecialSensors.equals(other.activeSpecialSensors)) {
```
ivansenic: as you removed the static method now these collections could be different, go for the `Collections.disjoint()` or make constructor package private?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationConfigurationChangedEvent.java
```diff
@@ -0,0 +1,33 @@
+package rocks.inspectit.oce.core.instrumentation.config;
```
ivansenic: move to `.event`

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationConfiguration.java
```diff
@@ -0,0 +1,19 @@
+package rocks.inspectit.oce.core.instrumentation.config;
```
ivansenic: move to `.model`

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/ClassInstrumentationConfiguration.java
```diff
@@ -0,0 +1,58 @@
+package rocks.inspectit.oce.core.instrumentation.config;
```
ivansenic: move to `.model`

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationConfigurationResolver.java
```diff
@@ -0,0 +1,118 @@
...
+        val config = currentConfig;
+        if (isIgnoredClass(clazz, config)) {
+            return ClassInstrumentationConfiguration.NO_INSTRUMENTATION;
+
+        } else {
+            TypeDescription description = TypeDescription.ForLoadedType.of(clazz);
+            Set<SpecialSensor> activeSensors = specialSensors.stream()
+                    .filter(s -> s.shouldInstrument(description, config))
+                    .collect(Collectors.toSet());
+            return new ClassInstrumentationConfiguration(activeSensors, config);
```
ivansenic: I have a feeling that for not ignored classes we are actually always creating the `ClassInstrumentationConfiguration` instance with an empty set of sensors.. this way we are always trying to instrument the class imo, just never actually change the byte code as a loop in the transformer is skipped.. We should return the `NO_INSTRUMENTATION` is there is no active sensor!

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/InstrumentationManager.java
ivansenic: let's just get them all from the `Instrumentation`

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/AsyncClassTransformer.java
```diff
@@ -0,0 +1,179 @@
...
+        instrumentation.removeTransformer(this);
+    }
+
+    private void deinstrumentAllClasses() {
+        while (!instrumentedClasses.isEmpty()) {
+            List<Class<?>> batchClasses = new ArrayList<>();
+            int maxBatchSize = env.getCurrentConfig().getInstrumentation().getInternal().getClassRetransformBatchSize();
+
+            Iterator<Class<?>> it = instrumentedClasses.iterator();
+            for (int i = 0; i < maxBatchSize && it.hasNext(); i++) {
```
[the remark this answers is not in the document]
JonasKunz: When doing a retransform a lock is held on all involved classloaders. Therefore, doing batches without sleeps has two benefits: 
(a) there are less classloaders locked at the same time, allowing the instrumented application with a higher probabiltiy to continue running
(b) between teh retransforms calls context switches are possible. This means that we don't prevent the isntrumented application from executing stuff for too long at once

I wouldn't use sleep here as it coudl greatly affect the shutdown speed, However, I will add a Thread.yield() instruction.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/NewClassDiscoveryService.java
```diff
@@ -0,0 +1,102 @@
...
+    @Autowired
+    private Instrumentation instrumentation;
+
+    @Autowired
+    private ScheduledExecutorService exec;
+
+    @Autowired
+    private List<IClassDiscoveryListener> listeners;
+
+    private Set<Class<?>> knownClasses = Collections.newSetFromMap(new WeakHashMap<>());
```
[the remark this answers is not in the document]
JonasKunz: Because we don't need concurrency here, this set is only modified by this class from a single thread.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/InstrumentationSettings.java
```diff
@@ -24,4 +25,10 @@
     @Valid
     private SpecialSettings special;
 
+    /**
+     * Defines which packages of the bootstrap should not be instrumented.
+     * All classes from the given packages and their subpackages will be ignored.
+     */
+    private Map<String, Boolean> ignoredBootstrapPackages;
```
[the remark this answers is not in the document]
JonasKunz: As discussed: Maps can be merged across different property sources, lists cannot.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/ClassInstrumentedEvent.java
```diff
@@ -0,0 +1,42 @@
...
+    @Getter
+    private final TypeDescription instrumentedClassDescription;
+
+    /**
+     * The instrumentation which has been applied.
+     * If no instrumentation was applied or the instrumentation has been removed this configuration
+     * is the same as {@link ClassInstrumentationConfiguration#NO_INSTRUMENTATION}.
+     */
+    @Getter
+    private final ClassInstrumentationConfiguration appliedConfiguration;
```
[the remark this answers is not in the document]
JonasKunz: Marius added the same comment :D

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/IClassDefinitionListener.java
```diff
@@ -0,0 +1,19 @@
+package rocks.inspectit.oce.core.instrumentation;
```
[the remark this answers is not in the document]
JonasKunz: Created an "events" which also includes spring events

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/ClassInstrumentationConfiguration.java
JonasKunz: Disjoint btw would be semantically incorrect! Disjoint only returns true, if they have not a single element in common. I added a simple implementation to CommonUtils

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationConfigurationResolver.java
JonasKunz: Solved by discussion: in case the class does not have any active sensor, a configuration is returned which "isSameAs" the NO_INSTRUMENTATION configuration.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/NewClassDiscoveryService.java
```diff
@@ -92,7 +81,7 @@
 
     @Override
     public void onNewClassDefined(String className, ClassLoader loader) {
-        lastNewClassDefinitionTimestamp = timestampMS.get();
+        numTrialsSinceLastClassDefinition.set(0L);
```
ivansenic: use `lazySet` pls...

PR #84, opened 2019-01-23
Closes #44 - Fixed tests, included error stack traces in console output.

### LINKED ISSUE #44: Fix occasional failure of system tests
The system tests occasionally fail because they try to read metrics from the OpenCensus API which are somehow not available. Strangely, the recording in inspectIT is executed prior to the tests, however it seems like the OpenCensus buffers the recorded data somewhere before it gets flushed and visible for the tests.

Currently we have added a sleep as a workaround, which however does not always work. We need to find a way to trigger the flushing of the metrics prior to performing test asertions. @ivansenic suggested that we could try to execute a synchronous get agains the prometheus endpoint. The hope is that this triggers the flushing of the metrics so that they are available for the tests.

JonasKunz:


inspectit-oce-agent/src/system-test/java/rocks/inspectit/oce/TestUtils.java
```diff
@@ -0,0 +1,32 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class TestUtils {
+
+    /**
+     * Opencensus internally manages a queue of events.
+     * We simply add a event to the queue and wait unti it is processed.
+     */
+    public static void waitForOpenCensusQueueToBeProcessed() {
+        AtomicBoolean entryProcessed = new AtomicBoolean(false);
```
ivansenic: why here both Atomic and sync? simply have atomic only and instead of wait/notify do a small sleep until you get true from the atomic?

inspectit-oce-agent/src/system-test/java/rocks/inspectit/oce/metrics/ClassLoaderMetricsSysTest.java
```diff
@@ -7,32 +7,36 @@
...
-
-        //ensure that the values are sane
-        long loadedVal = ((AggregationData.LastValueDataLong) loaded.getValue()).getLastValue();
-        long unloadedVal = ((AggregationData.LastValueDataLong) unloaded.getValue()).getLastValue();
-
-        assertThat(loadedVal).isLessThanOrEqualTo(ManagementFactory.getClassLoadingMXBean().getTotalLoadedClassCount());
-        assertThat(unloadedVal).isLessThanOrEqualTo(ManagementFactory.getClassLoadingMXBean().getUnloadedClassCount());
+    public void testClassLoaderMetricCapturing() {
+        await().atMost(5, TimeUnit.SECONDS).untilAsserted(() -> {
+            ViewData loadedData = viewManager.getView(View.Name.create("jvm/classes/loaded"));
```
ivansenic: should you also call here  `TestUtils.waitForOpenCensusQueueToBeProcessed();` first? At least you are ensuring that queue is drained before you check? Maybe actually outside the await()!

ivansenic: ahh ok this is fixed with:

```
@BeforeEach
    void flushMetrics() {
```

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/SpringConfiguration.java
```diff
@@ -1,27 +1,47 @@
...
             t.setName("inspectit-thread-" + threadCount.getAndIncrement());
             return t;
         });
+        return activeExecutor;
+    }
+
+    @PreDestroy
+    void destroy() {
+        if (activeExecutor != null) {
+            activeExecutor.shutdownNow();
```
ivansenic: not related to the test, but do we have a shutdown of the executor created in the context?

inspectit-oce-core/src/test/java/rocks/inspectit/oce/core/SpringTestBase.java
```diff
@@ -119,8 +119,9 @@ public void initialize(ConfigurableApplicationContext ctx) {
      * @param level the level to compare against.
      */
     public void assertNoLogsOfLevelOrGreater(Level level) {
+        //TODO: remove ignore for existing view errors as soon as we are able to unregister views in OpenCensus
         verify(env.mockAppender, times(0)).doAppend(argThat(
-                (le) -> le.getLevel().isGreaterOrEqual(level)));
+                (le) -> le.getLevel().isGreaterOrEqual(level) && !le.getMessage().contains("View with the name ")));
```
ivansenic: I would prefer that we log the other message on `INFO` and don't do this check?!

ivansenic: ahh I see level is parameter here.. then let's use Slf4j markers! you can create a marker that should be ignored and can be used in tests..

inspectit-oce-agent/src/system-test/java/rocks/inspectit/oce/metrics/GCMetricsSysTest.java
```diff
@@ -81,7 +81,7 @@ public void testGCPauseCapturing() throws Exception {
                 .findFirst().get();
 
 
-        assertThat(((AggregationData.SumDataLong) minorTime.getValue()).getSum()).isGreaterThan(0);
+        assertThat(((AggregationData.SumDataLong) minorTime.getValue()).getSum()).isGreaterThanOrEqualTo(0);
         assertThat(((AggregationData.SumDataLong) majorTime.getValue()).getSum()).isGreaterThan(0);
```
mariusoe: Can it happen, that we had only minor GCs but not a major one, so this would be 0? Or is the notification only sent on major ones?

inspectit-oce-agent/src/system-test/java/rocks/inspectit/oce/TestUtils.java
JonasKunz: Atomic is just used to have a settable boolean.

inspectit-oce-agent/src/system-test/java/rocks/inspectit/oce/metrics/GCMetricsSysTest.java
JonasKunz: I don't think we experienced this yet.. what I experienced here was that my minorgc was so fast that it had 0ms pause time.. that's why I adjusted the check

Issue #59, opened 2019-01-14
Extending configuration to enable an annotation based target selection


ivansenic:
Should support:

* interface
* super-class
* class-annotation
* method-annotation
* method parameters (by default all methods with given name)
* modifiers (only public by default? at least private should be not instrumented by default)

Please separate the class name, interface and super-class, we should allow configs that define all 3 and not like now only one FQN.

Nice to have (can be separated into further tickets and moved to another milestone):
* wild-card (this is great, but we need to be careful here for what we want a wild-card to be possible)
  * don't allow `*` only
* when the instrumenting by interface, instrument only methods available in that interface
* when the instrumenting by super-class, instrument only methods available in that super-class

mariusoe:
In #58, the configuration is basically covering all the points mentioned in this ticket except annotation based configuration, thus, it is renamed.

PR #97, opened 2019-01-31
 Closes #58 - Basic configuration for custom method instrumentation

### LINKED ISSUE #58: Basic method instrumentation config
* define our own common instrumentation property in the config, maybe extract to a separate file that's added in the same way to the property sources as the default config
* define basic stuff for supporting profiles, each profile should have a turn off property
* have a possibility to define simple method references (`className`, `methodName`, `interfaceName`)
* make sure that configurations for same method are merged
  * eaither solve this with `ClassResolver` or instrumentation callbacks
* it would be nice to have information in the combined config because of which instrumentation entry/entries is this instrumentation acitve (we were missing this in the old inspectIT)

For testing purposes we should be able to have a simple hook that is for now simply logging the combined configuration out. IF we implement the knowledge about where the instrumentation is coming from, we could print this in a trace level.

Testing should be more done on the integration level, as for the system test it's quite hard to test this.

mariusoe:


mariusoe:
This PR covers also #59 except for annotation matching

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationScopeResolver.java
```diff
@@ -0,0 +1,257 @@
...
+    private InstrumentationScope resolveScope(String key, InstrumentationScopeSettings scopeSettings) {
+        ElementMatcher.Junction<TypeDescription> typeMatcher = buildTypeMatcher(scopeSettings);
+        ElementMatcher.Junction<MethodDescription> methodMatcher = buildMethodMatcher(scopeSettings, typeMatcher);
+
+        if (typeMatcher == null) {
+            typeMatcher = any();
+            log.warn("|> The defined scope '{}' is not narrowing the type-scope, thus, matching ANY types! This may cause a negative performance impact!", key);
+        }
+
+        return new InstrumentationScope(typeMatcher, methodMatcher);
```
JonasKunz: shouldn't you use typeMatcher.and(declaresMethod(methodMatcher)) as a better Type Matcher?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationScopeResolver.java
```diff
@@ -0,0 +1,257 @@
...
+     * of String representing full qualified class names.
+     */
+    private ElementMatcher.Junction<MethodDescription> buildArgumentMatcher(String[] arguments) {
+        if (arguments == null) {
+            return null;
+        } else if (arguments.length == 0) {
+            return takesArguments(0);
+        } else {
+            MatcherBuilder<MethodDescription> builder = new MatcherBuilder<>();
+
```
JonasKunz: Here you actually match for a prefix. Imagine you specify two arguments. Some methods takes these two arguments but an additional one, this method is matched too.
Also you initially had an approach with a Map<Integer,String> to specify the argument filters, why did you go for the List in the end? the Map approach gives more flexibility and is not more difficult to write imo

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationScopeResolver.java
```diff
@@ -0,0 +1,257 @@
...
+     */
+    private ElementMatcher.Junction<TypeDescription> buildTypeMatcher(InstrumentationScopeSettings scopeSettings) {
+        if (scopeSettings.getTypeScope() == null) {
+            return null;
+        }
+
+        MatcherBuilder<TypeDescription> builder = new MatcherBuilder<>();
+
+        if (scopeSettings.getTypeScope().getSuperclass() != null) {
+            processSuperType(builder, scopeSettings.getTypeScope().getSuperclass());
```
JonasKunz: For this matcher you need an additional check to make sure it is a class.
And for the interfaces you need a check to ensure it is an interface. Otherwise you could match for interfaces on the superclass config and for classes on the interface config.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationScopeResolver.java
```diff
@@ -0,0 +1,257 @@
...
+     * Processing a single {@link MethodMatcherSettings} and adding it to the given {@link MatcherBuilder}.
+     */
+    private void processMethod(MatcherBuilder<MethodDescription> builder, MethodMatcherSettings matcherSettings) {
+        MatcherBuilder<MethodDescription> innerBuilder = new MatcherBuilder<>();
+
+        innerBuilder.and(matcherSettings.getIsConstructor(), isConstructor());
+        innerBuilder.and(buildVisibilityMatcher(matcherSettings.getVisibility()));
+        innerBuilder.and(buildArgumentMatcher(matcherSettings.getArguments()));
+
+        if (!matcherSettings.getIsConstructor()) {
```
JonasKunz: what happens if isConstructor is null? If you don't want to allow it add a @NotNull annotation to the config

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/model/ClassInstrumentationConfiguration.java
```diff
@@ -16,23 +17,27 @@
...
+    public static final ClassInstrumentationConfiguration NO_INSTRUMENTATION = new ClassInstrumentationConfiguration(Collections.emptySet(), Collections.emptySet(), null);
 
     /**
      * The sensors which are active for the target class.
      */
     @Getter
     private final Set<SpecialSensor> activeSpecialSensors;
 
+    @Getter
+    private Set<InstrumentationRule> activeRules;
```
JonasKunz: Here I would recommend storing signatures of matched methods instead

inspectit-oce-core/src/main/resources/config/default.yml
```diff
@@ -36,6 +36,51 @@ inspectit:
...
+#        # the targeted type - classes which should be instrumented have to match ALL of the specified elements within the type scope
+#        type-scope:
+#          # interfaces which have to be implemented
+#          interfaces:
+#            - name-pattern: "java.util.List"
+#              matcher-mode: "EQUALS_FULLY"
+#          # the superclass which has to be extended
+#          superclass:
+#            name-pattern: "java.util.AbstractList"
+#            matcher-mode: "EQUALS_FULLY"
```
JonasKunz: you can actually omit the EQUALS_FULLYs here, right?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/InstrumentationSettings.java
```diff
@@ -31,4 +33,15 @@
      */
     private Map<String, Boolean> ignoredBootstrapPackages;
 
+    /**
+     * The configuration of the defined scopes. The map's key represents an unique id for the related instrumentation scope.
+     */
+    @Valid
+    private Map<String, InstrumentationScopeSettings> scopes;
```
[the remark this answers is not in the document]
mariusoe: As discussed. This is separate because the scopes and DC (when we have data collectors) have a n:m relation to each other. Let's keep this for now and change if necessary.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/scope/AdvancedScopeSettings.java
```diff
@@ -0,0 +1,25 @@
...
+import lombok.EqualsAndHashCode;
+import lombok.NoArgsConstructor;
+
+/**
+ * Data container for advanced settings of an {@link InstrumentationScopeSettings}.
+ */
+@Data
+@NoArgsConstructor
+@EqualsAndHashCode
+public class AdvancedScopeSettings {
```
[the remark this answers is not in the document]
mariusoe: Just logical grouping. I thiunk we have more items in the future, thus, it could be nice to have them grouped.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/AsyncClassTransformer.java
```diff
@@ -155,6 +162,20 @@ private void deinstrumentAllClasses() {
...
+                // Apply the hook if necessary
+                for (InstrumentationRule rule : classConf.getActiveRules()) {
+                    if (log.isDebugEnabled()) {
+                        log.debug("Added hook to {} due to rule '{}'.", classBeingRedefined, rule.getName());
+                    }
+                    for (InstrumentationScope scope : rule.getScopes()) {
+                        if (log.isTraceEnabled()) {
+                            log.trace("|> {}", scope.getTypeMatcher());
+                        }
+                        builder = builder.visit(Advice.to(DispatchHookAdvice.class).on(scope.getMethodMatcher()));
```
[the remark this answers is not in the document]
mariusoe: Nice catch!

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationConfigurationResolver.java
```diff
@@ -73,10 +79,34 @@ public ClassInstrumentationConfiguration getClassInstrumentationConfiguration(Cl
...
+        return config.getRules().stream()
+                .map(rule -> Pair.of(
+                        rule.getName(),
+                        rule.getScopes()
+                                .stream()
+                                .filter(s -> s.getTypeMatcher().matches(typeDescription))
+                                .collect(Collectors.toSet())))
+                .filter(p -> !p.getRight().isEmpty())
+                .map(p -> new InstrumentationRule(p.getLeft(), p.getRight()))
+                .collect(Collectors.toSet());
```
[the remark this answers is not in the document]
mariusoe: As discussed, all matchers using ByteBuddy's ``HashCodeAndEqualsPlugin.Enhance`` annotation implementing ``equal`` and ``hashCode`` functions. I also added a test to ensure this.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationScopeResolver.java
```diff
@@ -0,0 +1,257 @@
...
+     * Creates the {@link InstrumentationScope} instances based on the {@link InstrumentationScopeSettings} contained in
+     * the given {@link InstrumentationSettings}.
+     *
+     * @param source the configuration used as basis for the {@link InstrumentationScope}s
+     * @return Returns a map containing {@link InstrumentationScope} instances. The keys are representing a unique identifier for each scope.
+     */
+    public Map<String, InstrumentationScope> resolve(InstrumentationSettings source) {
+        Map<String, InstrumentationScope> scopeMap = new HashMap<>();
+
+        if (source != null && source.getScopes() != null) {
```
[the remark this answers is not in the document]
mariusoe: Removed the source null check but I would keep the other one, because I think it is possible that it can be null - when no instrumentation is defined.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/model/ClassInstrumentationConfiguration.java
mariusoe: Related to the comment on equals and hashCode. Because these method should work as expected this should be fine.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/model/ClassInstrumentationConfiguration.java
```diff
@@ -50,6 +55,9 @@ public boolean isSameAs(TypeDescription forType, ClassInstrumentationConfigurati
         if (!CommonUtils.contentsEqual(activeSpecialSensors, other.activeSpecialSensors)) {
             return false;
         }
+        if (!CommonUtils.contentsEqual(getActiveRules(), other.getActiveRules())) {
```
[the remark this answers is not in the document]
mariusoe: Related to the comment on equals and hashCode. Because these method should work as expected this should be fine.

inspectit-oce-core/src/main/resources/config/default.yml
mariusoe: Yes, just as an example including all available attributes :)

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationScopeResolver.java
mariusoe: Changed to primitive type

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationScopeResolver.java
mariusoe: Done.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationScopeResolver.java
mariusoe: Done.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationScopeResolver.java
```diff
@@ -0,0 +1,257 @@
...
+            for (Map.Entry<String, InstrumentationScopeSettings> scopeEntry : source.getScopes().entrySet()) {
+                InstrumentationScopeSettings scopeSettings = scopeEntry.getValue();
+                if (scopeSettings.getTypeScope() == null) {
+                    log.warn("Processing of scope '{}' skipped. Empty type-scope is not allowed!", scopeEntry.getKey());
+                    continue;
```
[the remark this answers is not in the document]
mariusoe: Removed and and checked during config validation. A flag is added to enforce a wide scope.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationScopeResolver.java
mariusoe: Changed it back and added the NotNull annotation. By this the user can write "is-constructor" in the configuration otherwise, if it's a primitive, the config-settings is just named "constructor" which is less expressive and another pattern than the other boolean values in that config element

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/scope/TypeScope.java
```diff
@@ -28,8 +28,8 @@
     private NameMatcherSettings superclass;
 
     /**
-     * Matcher which have to match the class name.
+     * Matcher which have to match the type's name.
      */
-    private List<NameMatcherSettings> classes;
+    private List<NameMatcherSettings> types;
```
JonasKunz: Shouldn't it be "type" instead of "types"? this is why you didn't like "classes", right?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationScopeResolver.java
```diff
@@ -46,12 +42,16 @@
             for (Map.Entry<String, InstrumentationScopeSettings> scopeEntry : source.getScopes().entrySet()) {
                 InstrumentationScopeSettings scopeSettings = scopeEntry.getValue();
 
-                log.debug("Processing scope '{}'.", scopeEntry.getKey());
+                if (log.isDebugEnabled()) {
```
JonasKunz: I guess below the if check is there for performance reasons, but why is it required here?
"scopeEntry.getKey()" shouldn't really be that performance critical

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationScopeResolver.java
```diff
@@ -72,190 +72,108 @@ private InstrumentationScope resolveScope(InstrumentationScopeSettings scopeSett
...
-        }
-
-        List<AccessModifier> accessModifiers = Arrays.asList(modifiers);
-        boolean isPublic = accessModifiers.contains(AccessModifier.PUBLIC);
-        boolean isProtected = accessModifiers.contains(AccessModifier.PROTECTED);
-        boolean isPackage = accessModifiers.contains(AccessModifier.PACKAGE);
-        boolean isPrivate = accessModifiers.contains(AccessModifier.PRIVATE);
-        if (isPublic && isProtected && isPackage && isPrivate) {
-            return null;
+    private void processInterfaces(MatcherChainBuilder<TypeDescription> builder, NameMatcherSettings nameSettings) {
```
JonasKunz: More like "processInterface", isn't it? And also fix the comment

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationScopeResolver.java
```diff
@@ -72,190 +72,108 @@ private InstrumentationScope resolveScope(InstrumentationScopeSettings scopeSett
...
         }
 
-        builder.or(innerBuilder.build());
+        return builder.build();
     }
 
     /**
-     * Creating an {@link ElementMatcher} matching methods which takes arguments equals to the given array. The array consists
-     * of String representing full qualified class names.
+     * Processing the interface and superclass settings of the {@link TypeScope}.
```
JonasKunz: Fix the comment, you only check for the direct and indirect superclass here

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/util/MatcherChainBuilder.java
```diff
@@ -1,15 +1,16 @@
-package rocks.inspectit.oce.core.instrumentation.config;
+package rocks.inspectit.oce.core.instrumentation.config.util;
```
JonasKunz: I think you should put them into a package named "matchers" instead of "util"

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/util/SpecialElementMatchers.java
```diff
@@ -0,0 +1,135 @@
...
+        } else {
+            return null;
+        }
+    }
+
+    /**
+     * Creating an {@link ElementMatcher} matching methods which takes arguments equals to the given array. The array consists
+     * of String representing full qualified class names.
+     */
+    public static <T extends MethodDescription> ElementMatcher.Junction<T> argumentsAre(String[] arguments) {
```
JonasKunz: I think again that it is better to use a List instead of an array as argument here

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/util/SpecialElementMatchers.java
```diff
@@ -0,0 +1,135 @@
...
+
+            return builder.build();
+        }
+    }
+
+    /**
+     * Creating an {@link ElementMatcher} matching methods with the given access modifiers.
+     * Returns null in case the given array contains each modifier element at least once.
+     */
+    public static <T extends ModifierReviewable.OfByteCodeElement> ElementMatcher.Junction<T> visibilityIs(AccessModifier[] modifiers) {
```
JonasKunz: Same as above

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/scope/TypeScope.java
JonasKunz: Also: why is this a list while superclass is not a list? "Types" makes it sound like you can match for a disjunction of matchers while the NameMatchers are actually combined using conjunction.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationScopeResolver.java
```diff
@@ -222,16 +218,16 @@ private void processMethod(MatcherBuilder<MethodDescription> builder, MethodMatc
...
-            scopeSettings.getTypeScope().getInterfaces().forEach(i -> processSuperType(builder, i));
+            scopeSettings.getTypeScope().getInterfaces().forEach(i -> processInterfaces(builder, i));
         }
         if (scopeSettings.getTypeScope().getClasses() != null) {
             scopeSettings.getTypeScope().getClasses().forEach(c -> processClass(builder, c));
         }
 
-        if (scopeSettings.getAdvanced() != null && scopeSettings.getAdvanced().getInstrumentOnlyAbstractClasses()) {
+        if (scopeSettings.getAdvanced() != null && scopeSettings.getAdvanced().isInstrumentOnlyAbstractClasses()) {
             builder.and(isAbstract());
```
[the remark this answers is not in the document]
mariusoe: Ture. Don't know any usecase for this.. Remving it as discussed

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/scope/TypeScope.java
mariusoe: I think it does not makes sense to be able to define more than one superclass because all of them would have to match, so you can define the "lowest" one in that chain to achieve the same result.

Renamed the field into "type"

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationScopeResolver.java
mariusoe: I just didn't want to mix it up whinin one method :P

mariusoe: Removed it :)

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/scope/TypeScope.java
JonasKunz: What I mean: you could have something similar as for type, e.g. "The class must have a superclass from package xyz ending with ABC". Matching the superclass is conceptually almost the same as matching the type, therefore it makes sense to have a list of name matchers in either both or none.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/scope/MethodMatcherSettings.java
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Whether the method is a constructor or not.
+     */
+    @NotNull
+    private Boolean isConstructor = false;
+
+    /**
+     * Whether the method is synchronized or not. If this property is `null` the synchronize keyword will be ignored.
+     */
+    private Boolean isSynchronized;
```
ivansenic: At first, I was thinking this is really not necessary, but then I understood that it would be quite easy to instrument all synchronized methods with some wild cards, so quite handy..

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/scope/MethodMatcherSettings.java
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * The arguments which have to match the method's signature.
+     */
+    private List<String> arguments;
+
+    /**
+     * The methods visibility. On of the specified matcher has to match.
+     */
+    @NotNull
+    private List<AccessModifier> visibility = Arrays.asList(AccessModifier.PUBLIC, AccessModifier.PROTECTED, AccessModifier.PACKAGE, AccessModifier.PRIVATE);
```
ivansenic: definitely `PRIVATE` should not be included.. Actually, I am quite up to having only `PUBLIC` as default, decreasing the instrumentation size per default.. @JonasKunz your opinion?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/scope/TypeScope.java
mariusoe: As discussed, we'll cahnge it to a single matcher because using regex (yes it is more expensive) the same result can be achieved and the config is more consistent

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/scope/MethodMatcherSettings.java
mariusoe: Yes, good point. But I would rather go with just removing private from the list instead only providing public. But I'm also fine with just providing public by default.

JonasKunz: The two alternatives are (A) having either all as default or (B) only have public as default.
Benefit of (A): When the user gives a concrete name in the matcher, it works intuitively. E.g. he gives the name of a private method and he does not need to set anything in Addition to make it work as expected. The downside is if he uses a wildcard matcher he might instrument to many methods.
Benefit of (B): prevents issues when using a wildcard for method matchers whe nthere are very many classes and methods involved.

I think having all visibilities as default is fine ,because when do you usually use wildcard matchers? Mostly in combination with "instrument-only-inherited-methods", where the inherited methods are mostly public or protected anyway.

mariusoe: As discussed - keeping ALL as default

PR #94, opened 2019-01-29
Work on #61 - Added mechanism for generating generic data providers

JonasKunz:
This PR does not add any functionality visible for the end user, it only implemetns the logic required for building generic data providers.

PR #99, opened 2019-02-04
Closes #98 - Improved handling of retransformation failures

### LINKED ISSUE #98: Improve error handling in case of a retransformation exception
In some cases the retransformation can fail, e.g. if another Agent performs some incorrect bytecode changes. This causes the retransformClasses call to throw an exception ,which we already catch. However, we submit the classes to retransform in batches (default is 10 classes). When an exception occurs, no class from the batch is transformed, event the ones which would have worked correctly.
Therefore we need an additional failure handling: If a batch fails, we retry a retransform for each class individually. In addition we should expose a self-monitoring metric exposing the number of classes which failed to instrument.

JonasKunz:


JonasKunz:
Please review this PR after #94 has been integrated, as the test utilitites implemented tehre need to be used for updating the AsyncClassTransformer test.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/AsyncClassTransformer.java
```diff
@@ -130,8 +129,19 @@ private void deinstrumentAllClasses() {
...
+                if (batchClasses.size() > 1) {
+                    log.error("Error removing applied transformations for batch, retrying classes one by one", e);
+                    for (Class<?> clazz : batchClasses) {
+                        try {
+                            instrumentation.retransformClasses(clazz);
+                        } catch (Exception e2) {
+                            log.error("Error removing applied transformations of class '{}'", clazz.getName(), e2);
+                        }
+                    }
+                } else {
```
ivansenic: this will throw NPE if the batchClasses is empty, at least by looking into this part of the code it can happen..

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/AsyncClassTransformer.java
```diff
@@ -184,7 +194,8 @@ private void deinstrumentAllClasses() {
      * @param type                the classes type description
      * @return
      */
-    private ClassInstrumentationConfiguration updateAndGetActiveConfiguration(Class<?> classBeingRedefined, TypeDescription type) {
+    private ClassInstrumentationConfiguration updateAndGetActiveConfiguration
+    (Class<?> classBeingRedefined, TypeDescription type) {
```
ivansenic: what s going on here?

PR #100, opened 2019-02-05
Closes #96 - Added polling based alternative for detecting configuration directory changes

### LINKED ISSUE #96: Add a polling based alternative to the directory watching mechanisms
I found out that the configuration change notification of the config folder does not work in certain cases.
For example when the configuration folder is mounted from the network, it is possible that the Java WatchServices misses file change events.

I discovered this when using our demo on a Docker-toolbox installation. The config files get mounted perfectly and also changes are visible in the docker container, however no file system notifications are sent.

A simple solution would be to add a polling based WatchService alternative. We could use the following apache features for this: https://commons.apache.org/proper/commons-io/javadocs/api-release/index.html?org/apache/commons/io/monitor/package-summary.html
I would recommend that we add a "frequency" property to the inspectit.config.filebased configuration.
If the frequency is zero (which is the default), the WatchService is used. Otherwise we use a polling based approach.

JonasKunz:


inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/filebased/ConfigurationDirectoriesWatcher.java
```diff
@@ -24,8 +24,7 @@
  */
 @Service
 @Slf4j
-@ActivationConfigCondition("config.fileBased.watch")
-public class ConfigurationDirectoriesWatcher {
+public class ConfigurationDirectoriesWatcher extends DynamicallyActivatableService {
```
ivansenic: should we keep only one? do we have any benefit in having to implementations, or better said do we see any situation where the polling approach is not sufficient?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/filebased/ConfigurationDirectoriesPoller.java
```diff
@@ -0,0 +1,148 @@
...
+            pollingTask.cancel(true);
+            for (val poller : filePollers) {
+                try {
+                    poller.destroy();
+                } catch (Exception e) {
+                    log.error("Error destroying poller!", e);
+                }
+            }
+        }
+        return true;
```
ivansenic: clear file poolers list, when activated again you would have duplicated entries? ahh I see you will create a new list.. anyway, let's ensure this one is empty..

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/filebased/ConfigurationDirectoriesPoller.java
```diff
@@ -0,0 +1,148 @@
...
+
+    private class DirectoryPropertySourceChangeChecker {
+
+        private FileAlterationObserver observer;
+        private DirectoryPropertySource dirToWatch;
+        private boolean anyFileChanged = false;
+
+        public void init(DirectoryPropertySource dirToWatch) throws Exception {
+            this.dirToWatch = dirToWatch;
+            File path = dirToWatch.getRootDir().toFile();
```
ivansenic: you are not checking anywhere if this file exists? could this be a problem? or maybe I am missing the check?

JonasKunz: I have a try-catch there which would handle this case

PR #103, opened 2019-02-06
Closes #101 - Made system tests wait for the instrumentation to complete

### LINKED ISSUE #101: Make system tests wait for the instrumentation when required
Many system tests depend on the instrumentation being active. However, the instrumentation is performed asynchronously. Currenlty we simply wait for five seconds, which is fine on the ci but occassionally causes the system tests to fail when executing them locally.
To fix this we should make use of the self-monitoring of the instrumenter and wait until the instrumentationmanager queue is empty.

JonasKunz:


inspectit-oce-agent/src/system-test/java/rocks/inspectit/oce/TestUtils.java
```diff
@@ -20,5 +27,26 @@ public static void waitForOpenCensusQueueToBeProcessed() {
             throw new RuntimeException(e);
         }
     }
+
+    public static void waitForInstrumentationToComplete() {
+        await().atMost(30, TimeUnit.SECONDS).ignoreExceptions().untilAsserted(() -> {
+            assertThat(getInstrumentationQueueLength()).isZero();
```
mariusoe: Let's add also a check that we have at least one instrumented class.

inspectit-oce-agent/src/system-test/java/rocks/inspectit/oce/instrumentation/special/ExecutorContextPropagationTest.java
```diff
@@ -12,15 +12,15 @@
...
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static rocks.inspectit.oce.TestUtils.waitForInstrumentationToComplete;
 
 public class ExecutorContextPropagationTest {
 
     private static final Tagger tagger = Tags.getTagger();
 
     @BeforeAll
-    static void waitForInstrumentation() throws Exception {
```
mariusoe: Would it make sense that we create an abstract base class for all our system-tests which contains stuff like this (waiting for instrumentation)?
Then we just need to extend this class instead of having a separate before-all in each test.

mariusoe:
Please wait for merging of #105 :)

Issue #61, opened 2019-01-14
Generic hook uses data providers for key/value context pairs


ivansenic:
Create basic support for the data providers (collectors). 

* config model for the data providers
* data provider impl can be different
* they can collect data on entry, exit or both (have different methods for this of course)
* figure out how can data providers be referenced in the instrumentation config
* create/extend the hook, so that data providers are executed on the `entry` and `exit`
* if we have `before` and `after` methods instead of `proceed` ensure that the result object of before hook is passed to the after one

PR #104, opened 2019-02-06
Work on #61 - Added configuration for data providers

JonasKunz:


inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/InstrumentationSettings.java
```diff
@@ -21,31 +23,41 @@
...
+    @NotNull
+    private Map<@NotBlank String, @Valid GenericDataProviderSettings> dataProviders = Collections.emptyMap();
 
     /**
      * The configuration of the defined scopes. The map's key represents an unique id for the related instrumentation scope.
      */
     @Valid
     @NotNull
-    private Map<String, InstrumentationScopeSettings> scopes = Collections.emptyMap();
+    private Map<@NotBlank String, InstrumentationScopeSettings> scopes = Collections.emptyMap();
```
ivansenic: would we also need a Map<@NotBlank String, @Valid InstrumentationScopeSettings>to make sure that values are valid? same below?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/dataproviders/GenericDataProviderSettings.java
```diff
@@ -0,0 +1,82 @@
...
+     * In addition arbitrary custom input variables may be defined.
+     */
+    private Map<@NotBlank String, @NotBlank String> input = new HashMap<>();
+
+    /**
+     * A list of packages to import when compiling the java code and deriving the types of {@link #input}
+     * If a classname is not found, the given packages will be scanned in the given order to locate the class.
+     * This allows the User to use classes without the need to specify the FQN.
+     */
+    private List<@NotBlank String> imports = new ArrayList<>();
```
ivansenic: still regex for the package/FQN name would be great..

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/dataproviders/GenericDataProviderSettings.java
```diff
@@ -0,0 +1,82 @@
...
+    /**
+     * A string defining the Java method body of the data-provider without surrounding braces {}.
+     * This method body must have a return statement to return the value provided by the provider!
+     * The statement must be of type Object, primitive results have to wrapped manually!
+     * If this field is present, {@link #value} must be null!
+     */
+    private String valueBody;
+
+
+    @AssertTrue(message = "Either 'value' or 'valueBody' must be present (and not both)!")
```
ivansenic: `@AssertFalse` with `StringUtils.isEmpty(value) && StringUtils.isEmpty(valueBody)` would be way more readable ?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/DataProviderResolver.java
```diff
@@ -0,0 +1,81 @@
...
+     * Resolves {@link InstrumentationConfiguration#getDataProviders()}.
+     * Returns a map mapping the names of providers to their resolved configurations
+     *
+     * @param builder the builder to resolve
+     * @param source  the input configuration
+     * @return the updated builder
+     */
+    Map<String, ResolvedGenericDataProviderConfig> resolveProviders(InstrumentationSettings source) {
+        Map<String, ResolvedGenericDataProviderConfig> resultMap = new HashMap<>();
+        for (val entry : source.getDataProviders().entrySet()) {
```
ivansenic: `source.getDataProviders().entrySet().forEach(name, conf -> {...})`?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/DataProviderResolver.java
```diff
@@ -0,0 +1,81 @@
...
+        if (conf.getValue() != null) {
+            result.valueBody("return " + conf.getValue() + ";");
+        } else {
+            result.valueBody(conf.getValueBody());
+        }
+    }
+
+    private void resolveAdditionalInputs(HashMap<String, String> additionalInputs, ResolvedGenericDataProviderConfig.ResolvedGenericDataProviderConfigBuilder result) {
+        //everything that has not been removed is an additional argument
+        for (val additionalVarEntry : additionalInputs.entrySet()) {
```
ivansenic: `additionalInputs.entrySet().forEach(result::additionalArgumentType)` a bit more of Java8 style :-D

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/DataProviderResolver.java
```diff
@@ -0,0 +1,81 @@
...
+            result.additionalArgumentType(additionalVarEntry.getKey(), additionalVarEntry.getValue());
+        }
+    }
+
+    private void resolveArgumentVariables(HashMap<String, String> inputs, ResolvedGenericDataProviderConfig.ResolvedGenericDataProviderConfigBuilder result) {
+        val entryIterator = inputs.entrySet().iterator();
+        while (entryIterator.hasNext()) {
+            val varEntry = entryIterator.next();
+            String varName = varEntry.getKey();
+            if (varName.matches(DataProviderGenerator.ARG_VARIABLE_PREFIX + "\\d+")) {
```
ivansenic: not sure if we should use the greedy variant `\\d*` which will only match if rest of the string are numbers..

this way you would match: `arg1 and_antyhing_else ` and you would then fail with the `parseInt`.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/dataprovider/generic/DataProviderGenerator.java
```diff
@@ -168,11 +168,12 @@ private String buildProviderMethod(ResolvedGenericDataProviderConfig providerCon
             buildVariableDefinition(methodBody, type, varName, value);
         });
         val additionalArgs = providerConfig.getAdditionalArgumentTypes();
+        val iterator = additionalArgs.entrySet().iterator();
         for (int id = 0; id < additionalArgs.size(); id++) {
```
ivansenic: should we then have a loop that corresponds with the iterator style?

inspectit-oce-core/src/main/resources/config/default.yml
```diff
@@ -107,6 +107,21 @@ inspectit:
       # defines how often after the last invocation of a ClassFileTransformer the agent scans for new classes
       num-class-discovery-trials: 2
 
+    #definition of commonly used data providers
+    data-providers:
+
+      #computes a nanosecond-timestamp as a long for the current point in time
+      timestampNanos:
+        value: "new Long(System.nanoTime())"
```
ivansenic: So we always need to work with objects right? that's because we are returning the `Object` as the result of the data providers.. I am only thinking that we are making a lot of garbage this way, but anyway we would be wrapping this into some kind of map or list, thus we would anyway need an object.. Right?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/InstrumentationSettings.java
JonasKunz: We already have valid as an annotation on the attirbute there, this propagates to the type parameter. But iI guess it is more readable to have it explicitly there.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/dataproviders/GenericDataProviderSettings.java
JonasKunz: We don't only check if both are empty but also check that not both are given, I will separate it into two checks

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/DataProviderResolver.java
JonasKunz: This one and the comment above are left-overs from where I had to write to a local variable outside of the loop :-D

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/DataProviderResolver.java
JonasKunz: String.matches matches the entire string, not only a prefix. This means for your example it  would not match the pattern.

inspectit-oce-core/src/main/resources/config/default.yml
JonasKunz: Yes we store everything in a map, that's why we need an object anyway. To my knowledge constructing small objects with a short livetime isn't that much of a problem in modern JVMs anymore..

inspectit-oce-core/src/test/java/rocks/inspectit/oce/core/instrumentation/config/DataProviderResolverUnitTest.java
```diff
@@ -0,0 +1,197 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class DataProviderResolverUnitTest {
+
+    @InjectMocks
+    private DataProviderResolver resolver = new DataProviderResolver();
+
+    @Nested
+    public class ResolveDataProviders {
```
mariusoe: Rename to `ResolveProviders`

inspectit-oce-core/src/test/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationConfigurationResolverUnitTest.java
```diff
@@ -27,7 +27,7 @@
 import static org.mockito.Mockito.when;
 
 @ExtendWith(MockitoExtension.class)
-class InstrumentationConfigurationResolverTest {
+class InstrumentationConfigurationResolverUnitTest {
```
mariusoe: Please keep unit tests named just `<ClassName>Test` and name all tests which are not unit tests differently, e.g. as we already did with `<ClassName>IntTest`

inspectit-oce-core/src/test/java/rocks/inspectit/oce/core/instrumentation/config/DataProviderResolverUnitTest.java
```diff
@@ -0,0 +1,197 @@
...
+        @BeforeEach
+        void setup() {
+            config = new InstrumentationSettings();
+            inputProvider = new GenericDataProviderSettings();
+            inputProvider.setValueBody(PROVIDER_VALUE_BODY);
+            config.setDataProviders(Maps.newHashMap(PROVIDER_NAME, inputProvider));
+        }
+
+        @Test
+        void verifyCaseSensitivity() {
```
mariusoe: I think I don't get the name of this test. Are you testing some casing stuff in this test? I mean, you basically just check if the name is correctly resolved, or do I miss something?

JonasKunz: I think I can jsut remove this test. This was a left over when I had the stuff case insensitive

PR #107, opened 2019-02-08
Work on #61 - Added context sensitive config validation mechanism and configuration for data provider calls

JonasKunz:


inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/InstrumentationSettings.java
```diff
@@ -58,4 +62,11 @@
      */
     @NotNull
     private Map<@NotBlank String, @Valid InstrumentationRuleSettings> rules = Collections.emptyMap();
+
+    @AdditionalValidation
```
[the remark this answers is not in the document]
JonasKunz: Solved by discussion: AssertTrue does not allow you to parametrize the error messages. For example for a list of scopes you would not be able to output which of the given scopes does not exist.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/dataproviders/DataProviderCallSettings.java
```diff
@@ -0,0 +1,172 @@
...
+    @NotBlank
+    private String provider;
+
+    /**
+     * Defines input values for the provider using constants, e.g. numbers or string literals.
+     * The key of the map denotes the input variable of the provider, the value is the constant value to assign.
+     * The given string literals are converted to the corresponding primitive type or directly used as a string.
+     * Constant input only works for inputs which have a primitive type, a corresponding wrapper type or are strings.
+     */
+    private Map<@NotBlank String, Object> constantInput = Collections.emptyMap();
```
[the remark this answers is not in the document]
ivansenic: OK I see tests are clearly stating this works :+1:

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/dataproviders/GenericDataProviderSettings.java
```diff
@@ -46,7 +56,8 @@
      * If a classname is not found, the given packages will be scanned in the given order to locate the class.
      * This allows the User to use classes without the need to specify the FQN.
      */
-    private List<@Pattern(regexp = PACKAGE_REGEX) String> imports = new ArrayList<>();
+    private List<@javax.validation.constraints.Pattern(regexp = PACKAGE_REGEX)
```
ivansenic: :dancer:

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/rules/InstrumentationRuleSettings.java
```diff
@@ -25,7 +31,48 @@
...
+    @NotNull
+    private Map<@NotBlank String, @NotNull @Valid DataProviderCallSettings> exitData = Collections.emptyMap();
+
+    /**
+     * Validates this rule, invoked by {@link InstrumentationSettings#performValidation(ViolationBuilder)}
+     *
+     * @param container the settings containing this rule
+     * @param vios      the violation builder
+     */
+    public void performValidation(InstrumentationSettings container, ViolationBuilder vios) {
```
ivansenic: can we add a unit test for this as well please..

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/validation/AdditionalValidationsValidator.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+import static java.util.stream.Collectors.toList;
+
+/**
+ * Hibernate validator class for the {@link AdditionalValidations} annotation.
+ */
+public class AdditionalValidationsValidator implements ConstraintValidator<AdditionalValidations, Object> {
```
ivansenic: Unit test this as well, please..

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/utils/CommonUtils.java
```diff
@@ -32,4 +34,28 @@ public static boolean contentsEqual(Set<?> setA, Set<?> setB) {
...
+     * Tries to find the given type in the given classloader or one of its parents.
+     * Hereby, the given imported packages are taken into account.
+     * In addition, java.lang is assumed to be an implicit import.
+     *
+     * @param typename the name of the type, can also be fully qualified
+     * @param context  the classloader to search in
+     * @param packages the imported packages, e.g. "java.util", "javax.servlet"
+     * @return the Class if it was found, null otherwise
+     */
+    public static Class<?> locateTypeWithinImports(String typename, ClassLoader context, Collection<String> packages) {
```
ivansenic: nice solution.. I guess you used `Stream.concat` to preserve order of trying:

1. FQN
2. something with declared import
3. Java class

I guess this is fine imo, can you just format a bit code in this method so that this is more readable:

```
return Stream.concat(
                Stream.concat(
                    Stream.of(""),
                    packages.stream().map(s -> s + ".")
                ),
                Stream.of("java.lang.")
            )
            .flatMap(prefix -> {
                try {
                    return Stream.of(Class.forName(prefix + typename, false, context));
                } catch (Exception e) {
                    return Stream.empty();
                }
            }).findFirst().orElse(null);
```

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/dataproviders/DataProviderCallSettings.java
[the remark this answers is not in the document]
JonasKunz: When using the inputs from this map we just have to pipe them through an ApplicationConversionService and everything will work because of spring :)
The conversion service can also be called with a generic signature as future work, so that e.g. you could pass Lists of Durations and something like this in.

PR #108, opened 2019-02-08
Closes #59 - Extending configuration to enable an annotation based target selection

### LINKED ISSUE #59: Extending configuration to enable an annotation based target selection

mariusoe:


inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationScopeResolver.java
```diff
@@ -94,33 +95,31 @@ private InstrumentationScope resolveScope(InstrumentationScopeSettings scopeSett
...
-     * Creating a matcher for the given {@link NameMatcherSettings} which represents concrete classes.
+     * Creating and adding a matcher for the given {@link NameMatcherSettings} which represents concrete classes.
      */
-    private void processType(MatcherChainBuilder<TypeDescription> builder, NameMatcherSettings nameSettings) {
-        ElementMatcher.Junction<TypeDescription> nameMatcher = nameIs(nameSettings);
-        if (nameMatcher != null) {
-            builder.and(nameMatcher);
-        }
+    private void processType(MatcherChainBuilder<TypeDescription> builder, DescriptionMatcherSettings descriptionSettings) {
+        ElementMatcher.Junction<TypeDescription> nameMatcher = describedBy(descriptionSettings);
```
JonasKunz: Maybe also rename the variable "nameMatcher" here, as it does not only match by name anymore

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/hook/DispatchHookAdvice.java
```diff
@@ -14,8 +14,8 @@
      * The content of this method is added before the target method! See {@link Advice} for existing parameter annotations.
      */
     @Advice.OnMethodEnter
-    public static void onEnter(@Advice.Origin Object origin) {
-        System.out.println("\tEntering: " + origin);
+    public static void onEnter(@Advice.Origin Object origin, @Advice.This Object thiz) {
```
JonasKunz: this still also works on static methods, right?

inspectit-oce-core/src/test/java/rocks/inspectit/oce/core/config/model/instrumentation/scope/DescriptionMatcherSettingsTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+        @Test
+        public void allAnnotationMatcher() {
+            DescriptionMatcherSettings settings = new DescriptionMatcherSettings();
+            settings.setMatcherMode(StringMatcher.Mode.STARTS_WITH);
+            NameMatcherSettings matcher = new NameMatcherSettings();
+            matcher.setMatcherMode(StringMatcher.Mode.STARTS_WITH);
+            settings.setAnnotations(Collections.singletonList(matcher));
+
+            boolean result = settings.isAnyMatcher();
```
JonasKunz: why is this an any-matcher? It only matches types which have at least one annotation, doesn't it?

.gitignore
```diff
@@ -1,6 +1,7 @@
 .gradle
 *build/
 !gradle/wrapper/gradle-wrapper.jar
+*.log
```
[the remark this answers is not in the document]
mariusoe: Yep :)

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/scope/DescriptionMatcherSettings.java
```diff
@@ -0,0 +1,32 @@
...
+import java.util.List;
+
+/**
+ * Extending the {@link NameMatcherSettings} by adding the ability to specify a list of {@link NameMatcherSettings} which
+ * represents annotations.
+ */
+@Data
+@NoArgsConstructor
+@EqualsAndHashCode(callSuper = true)
+public class DescriptionMatcherSettings extends NameMatcherSettings {
```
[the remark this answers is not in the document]
mariusoe: That was the best name I had :D but I'll go for the `ElementDescriptionMatcherSettings` your proposed

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationScopeResolver.java
```diff
@@ -94,33 +95,31 @@ private InstrumentationScope resolveScope(InstrumentationScopeSettings scopeSett
...
     /**
-     * Creating a matcher for the given {@link NameMatcherSettings} which represents a superclass.
+     * Creating and adding a matcher for the given {@link NameMatcherSettings} which represents a superclass.
      */
-    private void processSuperclass(MatcherChainBuilder<TypeDescription> builder, NameMatcherSettings nameSettings) {
-        ElementMatcher.Junction<NamedElement> nameMatcher = nameIs(nameSettings);
-        if (nameMatcher != null) {
-            builder.and(hasSuperType(not(isInterface()).and(nameMatcher)));
+    private void processSuperclass(MatcherChainBuilder<TypeDescription> builder, DescriptionMatcherSettings descriptionSettings) {
+        ElementMatcher.Junction<TypeDescription> matcher = describedBy(descriptionSettings);
```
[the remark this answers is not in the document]
mariusoe: Using `ElementDescriptionMatcherSettings`

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/hook/DispatchHookAdvice.java
mariusoe: No it doesn't :) this was a test and shouldn't be checked in

inspectit-oce-core/src/test/java/rocks/inspectit/oce/core/config/model/instrumentation/scope/DescriptionMatcherSettingsTest.java
mariusoe: You're right. This case shouldn't be a problem - changed it.

PR #109, opened 2019-02-08
Closes #95 - Implemented InspectitContext for passing data around

### LINKED ISSUE #95: Implement a custom GRPC context value for passing data
OpenCensus uses GRPC to pass TagContexts around. We however see our "data" passed around as a superset of the tags. For example, the data is not limited to being a String type, data can be arbitrary objects. For this reason we need to implement a custom Context, which however keeps the OpenCensus context in sync.

JonasKunz:
This PullRequest was opened so that you could review my idea on how we should handle the passing of data around within the JVM.

This context is built to work also with TagContexts opened by the user in the instrumented application.
E.g. if the User creates a new tag within a trace or overwrite a tags value, this is visible for the next InspectITContext which gets opened in the trace.


inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/context/InspectitContext.java
```diff
@@ -0,0 +1,168 @@
...
+import rocks.inspectit.oce.core.instrumentation.config.model.ResolvedDataProperties;
+import rocks.inspectit.oce.core.tags.CommonTagsManager;
+
+import java.util.*;
+
+public class InspectitContext implements AutoCloseable {
+
+    /**
+     * Package private for testing
+     */
```
mariusoe: Guava provides and annotation for this `@VisibleForTesting`. It is doing nothing but just for marking an method that the visibility is wider than required, so you don't have to comment this in the docs.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/context/InspectitContext.java
```diff
@@ -0,0 +1,168 @@
...
+
+import io.grpc.Context;
+import io.opencensus.common.Scope;
+import io.opencensus.tags.*;
+import rocks.inspectit.oce.core.instrumentation.config.model.ResolvedDataProperties;
+import rocks.inspectit.oce.core.tags.CommonTagsManager;
+
+import java.util.*;
+
+public class InspectitContext implements AutoCloseable {
```
mariusoe: I'm lost in this class :P Please let's discuss this class in a call

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/context/ContextManagerImpl.java
```diff
@@ -20,24 +23,43 @@
...
+    public ContextManagerImpl(CommonTagsManager commonTags, InstrumentationConfigurationResolver configProvider) {
+        this.commonTags = commonTags;
+        this.configProvider = configProvider;
+    }
+
     @Override
     public Runnable wrap(Runnable r) {
-        return Context.current().wrap(r);
+        Runnable result;
+        try (val upPropagationBarrier = enterNewContext()) {
```
ivansenic: I am still confused about why we need to open a new context every time we switch the threads.. I guess this is there to purely ensure that everything added by the user to the OC tag contact is combined with our latest context and propagated together. But in fact, both will be propagated in different GRPC keys, so this for me is a huge overhead.. Let's sync on this..

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/context/InspectitContext.java
```diff
@@ -0,0 +1,168 @@
...
+        if (parent == null) {
+            result.data.putAll(commonTags.getCommonTagValueMap());
+        } else {
+            parent.data.entrySet().stream()
+                    .filter(e -> dataProperties.isPropagatedDownWithinJVM(e.getKey()))
+                    .forEach(e -> result.data.put(e.getKey(), e.getValue()));
+        }
+
+        if (parentTags != null) {
+            //we copy the data from any opened tag context in case it was not opened by the inspectit parent context
```
ivansenic: let's talk about this part as well

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/data/DataSettings.java
```diff
@@ -0,0 +1,40 @@
...
+     * Defines how data is propagation up within traces.
+     * Defaults to {@link PropagationMode#NONE} when not specified.
+     * The default values handling is done by {@link rocks.inspectit.oce.core.instrumentation.config.DataPropertiesResolver}
+     * as well through the qzery methods of {@link ResolvedDataProperties}.
+     */
+    private PropagationMode upPropagation;
+
+    /**
+     * Defines how data is propagation down within traces.
+     * Defaults to {@link PropagationMode#NONE} if the name of the data starts with "local_", otherwise defautls to {@link PropagationMode#JVM_LOCAL}.
```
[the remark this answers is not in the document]
JonasKunz: Agreed :)

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/data/DataSettings.java
```diff
@@ -0,0 +1,40 @@
...
+import rocks.inspectit.oce.core.config.model.instrumentation.InstrumentationSettings;
+import rocks.inspectit.oce.core.instrumentation.config.model.ResolvedDataProperties;
+
+/**
+ * Defines the behaviour of data with a certain key passed around with the {@link rocks.inspectit.oce.core.instrumentation.context.InspectitContext}
+ * for instrumented methods. The data key is defined through the map {@link InstrumentationSettings#getData()}.
+ */
+@Data
+@NoArgsConstructor
+public class DataSettings {
```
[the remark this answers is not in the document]
JonasKunz: Is included in the Data annotation

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/InstrumentationSettings.java
```diff
@@ -58,4 +62,22 @@
      */
     @NotNull
     private Map<@NotBlank String, @Valid InstrumentationRuleSettings> rules = Collections.emptyMap();
+
+    /**
+     * Defines the behaviour of the data regarding context propagation.
+     * E.g. is data propagated up and/or down, is it visible as a Tag for metrics collection?
+     */
+    @NotNull
+    private Map<@NotBlank String, @Valid DataSettings> data = Collections.emptyMap();
```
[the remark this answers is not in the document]
JonasKunz: I think spring never puts null values into a map, tehrefore no NotNull is needed here

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/context/ContextManagerImpl.java
```diff
@@ -20,24 +23,43 @@
...
+    private CommonTagsManager commonTags;
+
+    private InstrumentationConfigurationResolver configProvider;
+
     /**
      * Cache for storing the context objects.
      */
     private final Cache<Thread, Context> storedContexts = CacheBuilder.newBuilder().weakKeys().build();
 
+    public ContextManagerImpl(CommonTagsManager commonTags, InstrumentationConfigurationResolver configProvider) {
```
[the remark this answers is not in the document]
JonasKunz: The reason is that spring tries to scan for the inherited interfaces, which it doesn't find because it is on the bootstrap.

inspectit-oce-core/src/test/java/rocks/inspectit/oce/core/instrumentation/context/InspectitContextUnitTest.java
```diff
@@ -0,0 +1,748 @@
...
+
+
+        @Test
+        void verifyDownPropagationForAsyncChildrenOnDifferentThreadWithRootActive() throws Exception {
+            when(commonTags.getCommonTagValueMap()).thenReturn(Collections.emptyMap());
+            doReturn(true).when(propagation).isPropagatedDownWithinJVM(any());
+
+            InspectitContext root = InspectitContext.createFromCurrent(commonTags, propagation, false);
+            root.setData("tag", "rootValue");
+            root.makeActive(false);
```
ivansenic: when would we use this situation actually? is this `boolean` in the active only for testing purposes, if so you can have that on the package level and introduce `makeActive()` that calls it with `true`

inspectit-oce-core/src/test/java/rocks/inspectit/oce/core/instrumentation/context/InspectitContextUnitTest.java
```diff
@@ -0,0 +1,748 @@
...
+
+            asyncTask.start();
+            asyncTask.join();
+
+            assertThat(tagValue.get()).isEqualTo("rootValue");
+        }
+
+
+        @Test
+        void verifyDownPropagationForAsyncChildrenOnSameThread() throws Exception {
```
ivansenic: again not async :)

inspectit-oce-core/src/test/java/rocks/inspectit/oce/core/instrumentation/context/InspectitContextUnitTest.java
```diff
@@ -0,0 +1,748 @@
...
+        void verifyCommonTagsExtracted() {
+            HashMap<String, String> tags = new HashMap<>();
+            tags.put("tagA", "valueA");
+            tags.put("tagB", "valueB");
+            when(commonTags.getCommonTagValueMap()).thenReturn(tags);
+
+            InspectitContext ctx = InspectitContext.createFromCurrent(commonTags, propagation, true);
+            ctx.makeActive(true);
+
+            assertThat(ctx.getData("tagA")).isEqualTo("valueA");
```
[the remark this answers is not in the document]
ivansenic: you are actually not testing anything with the OC context in the `DownPropagation` test set.. There is also no test for OC context opening between two inspectIT contexts

inspectit-oce-core/src/test/java/rocks/inspectit/oce/core/instrumentation/context/InspectitContextUnitTest.java
```diff
@@ -0,0 +1,748 @@
...
+
+
+        @Test
+        void verifyNoUpPropagationForAsyncChildrenOnDifferentThreadWithRootActive() throws Exception {
+            when(commonTags.getCommonTagValueMap()).thenReturn(Collections.emptyMap());
+            lenient().doReturn(true).when(propagation).isPropagatedUpWithinJVM(any());
+
+            InspectitContext root = InspectitContext.createFromCurrent(commonTags, propagation, false);
+            root.setData("tag", "rootValue");
+            root.makeActive(false);
```
ivansenic: why are you always calling `false` when testing "async" stuff, this is only related to the OC context, shouldn't `true` here also work fine..

inspectit-oce-core/src/test/java/rocks/inspectit/oce/core/instrumentation/context/InspectitContextUnitTest.java
```diff
@@ -0,0 +1,748 @@
...
+        }
+
+
+        @Test
+        void verifyCommonTagsPublished() {
+            HashMap<String, String> tags = new HashMap<>();
+            tags.put("tagA", "valueA");
+            tags.put("tagB", "valueB");
+            when(commonTags.getCommonTagValueMap()).thenReturn(tags);
+            doReturn(true).when(propagation).isPropagatedDownWithinJVM(any());
```
ivansenic: shouldn't the mocking of the propagation be removed in this test? or actually return `false` for both

inspectit-oce-core/src/test/java/rocks/inspectit/oce/core/instrumentation/context/InspectitContextUnitTest.java
```diff
@@ -0,0 +1,748 @@
...
+            InspectitContext ctx = InspectitContext.createFromCurrent(commonTags, propagation, true);
+            ctx.setData("local", "localValue");
+            ctx.setData("global", "globalValue");
+
+            ctx.makeActive(true);
+
+            assertThat(getCurrentTagsAsMap()).hasSize(1);
+            assertThat(getCurrentTagsAsMap()).containsEntry("global", "globalValue");
+
+            try (Scope scope = ctx.enterNewTagScope()) {
```
ivansenic: don't get this at all?

inspectit-oce-core/src/test/java/rocks/inspectit/oce/core/instrumentation/context/InspectitContextUnitTest.java
```diff
@@ -0,0 +1,748 @@
...
+            assertThat(getCurrentTagsAsMap()).containsEntry("rootKey1", "rootValue1");
+            assertThat(getCurrentTagsAsMap()).containsEntry("rootKey2", "rootValue2");
+
+            InspectitContext nested = InspectitContext.createFromCurrent(commonTags, propagation, true);
+            nested.makeActive(true);
+            nested.setData("rootKey1", "nestedValue1");
+            nested.setData("nestedKey2", "nestedValue2");
+            nested.close();
+
+            assertThat(getCurrentTagsAsMap()).hasSize(2);
```
ivansenic: what the heck how is this now working with no propagation?

JonasKunz: I will add a verification that up-propagated values are available in the full scope.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationConfigurationResolver.java
```diff
@@ -150,9 +155,19 @@ private void updateConfiguration(InstrumentationSettings source) {
...
                 .dataProviders(dataProviderResolver.resolveProviders(source))
+                .dataProperties(resolveDataProperties(source))
                 .build();
     }
 
+    private ResolvedDataProperties resolveDataProperties(InstrumentationSettings source) {
+        val defaultSettings = new DataSettings();
+        val builder = ResolvedDataProperties.builder();
+        source.getData().entrySet().stream()
+                .filter(e -> !defaultSettings.equals(e.getValue()))
```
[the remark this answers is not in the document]
JonasKunz: Actually discovered that this is not necessary, as data with the default settings will not end up in any of the sets of ResolvedDataProperteis anyway.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/context/InspectitContext.java
```diff
@@ -0,0 +1,409 @@
...
+     * <p>
+     * At the end of the entry phase, the map is the same as {@link #postEntryPhaseDownPropagatedData}
+     * When now an up-propagation occurs, this map becomes stale. Therefore it is "reset" to null and recomputed when it is required.
+     * <p>
+     * Note that the underlying map never gets altered! It gets replaced by a new object when it became stale.
+     * This ensures that child context can use this map as their {@link #postEntryPhaseDownPropagatedData} without copying!
+     */
+    private Map<String, Object> cachedActivePhaseDownPropagatedData = null;
+
+    private final ResolvedDataProperties propagation;
```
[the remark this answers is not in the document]
JonasKunz: parent cannot be made final because we have to clear it at the call of close()...

inspectit-oce-core/src/test/java/rocks/inspectit/oce/core/instrumentation/config/DataPropertiesResolverTest.java
```diff
@@ -0,0 +1,152 @@
...
+            assertThat(dataProps.isPropagatedDownGlobally("my_key")).isFalse();
+
+            assertThat(dataProps.isPropagatedUpWithinJVM("my_key")).isFalse();
+            assertThat(dataProps.isPropagatedUpGlobally("my_key")).isFalse();
+
+            assertThat(dataProps.isTag("my_key")).isTrue();
+        }
+
+        @Test
+        void defaultSettingsForUnmentionedLocalKeyCorrect() {
```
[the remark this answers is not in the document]
JonasKunz: Now I see why you foud nthis tests :D the file is outdated.
It was deleted and the relevant tests have been moved to the InstrumentationConfigurationResolverTest.

inspectit-oce-core/src/test/java/rocks/inspectit/oce/core/instrumentation/context/InspectitContextUnitTest.java
```diff
@@ -0,0 +1,748 @@
...
+
+    @Mock
+    CommonTagsManager commonTags;
+
+    @Mock
+    ResolvedDataProperties propagation;
+
+    Map<String, String> getCurrentTagsAsMap() {
+        HashMap<String, String> result = new HashMap<>();
+        InternalUtils.getTags(Tags.getTagger().getCurrentTagContext())
```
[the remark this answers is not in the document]
JonasKunz: Doesn't work because InternalUtils.getTags returns an Iterator and not an Iterable, therefore it is not trivial to create a stream out of it.

inspectit-oce-core/src/test/java/rocks/inspectit/oce/core/instrumentation/context/InspectitContextUnitTest.java
JonasKunz: The mocking just ensures the default behaviour of the data configuration. This is needed for makeActive to know that it should publish its data, which includes the common tags as a tag scope.

PR #110, opened 2019-02-11
Closes #79 - Added user excludes with intellij excluded by default

### LINKED ISSUE #79: User defined exclude packages/classes
The user should have possibility to exclude classes by package or FQN.

JonasKunz:


inspectit-oce-core/src/main/resources/config/default.yml
```diff
@@ -133,6 +133,13 @@ inspectit:
       '[com.google.]': true
       '[java.lang.invoke.]': true
 
+    # defines all packages packages whose classes and subpackages should be ignored
+    # the given strings are matches as prefixes for classes, therefore packages have to end with a dot
+    # however, for the trailing dots to be interpreted correctly you need to enclose the package in '[<package>]'
+    ignored-packages:
```
mariusoe: Basically, you match the classes and not the packages right. I mean you can also ignore single classes instead of package - thus this name is basically not 100% correct.
nevertheless, I also don't know a better naming for this property and I think this is totally fine. So, this comment is basically useless :)

JonasKunz: Solved by discussion: we keep this name but we have to clarify it in the documentation

PR #113, opened 2019-02-20
Closes #61 - Implemented binding and execution of data providers

### LINKED ISSUE #61: Generic hook uses data providers for key/value context pairs

JonasKunz:


inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/MethodHookConfigurationResolver.java
```diff
@@ -1,23 +1,186 @@
...
+import rocks.inspectit.oce.core.instrumentation.config.model.ResolvedDataProviderCall;
+import rocks.inspectit.oce.core.instrumentation.dataprovider.generic.DataProviderGenerator;
 
-import java.util.Set;
+import java.util.*;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
 @Component
 public class MethodHookConfigurationResolver {
```
ivansenic: I did really my best to understand this, but you need to add more explanations in the code it self, there's so much going on here, that this is quite hard to understand.. I know that the end result will be, but how exactly was it reached is not clear..

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/dataprovider/generic/BoundDataProvider.java
```diff
@@ -0,0 +1,163 @@
...
+            return new ConstantOnlyBoundDataProvider(dataKey, providerConfig, provider, constantAssignments);
+        } else {
+            return new DynamicBoundDataProvider(dataKey, providerConfig, provider, constantAssignments, dynamicAssignments);
+        }
+    }
+}
+
+class ConstantOnlyBoundDataProvider extends BoundDataProvider {
+
+    private Object[] arguments;
```
ivansenic: this should also be final, as well as the two fields in the dynamic data provider.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/dataprovider/generic/BoundDataProvider.java
```diff
@@ -0,0 +1,163 @@
...
+        }
+        this.dynamicAssignments = dynamicAssignmentsWithIndices.toArray(new Pair[0]);
+    }
+
+    @Override
+    public void execute(ExecutionContext context) {
+        Object[] args = Arrays.copyOf(argumentsTemplate, argumentsTemplate.length);
+
+        for (val assignment : dynamicAssignments) {
+            args[assignment.getLeft()] = assignment.getRight().apply(context);
```
ivansenic: let's discuss here shortly

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/hook/HookManager.java
```diff
@@ -182,44 +160,49 @@ private void checkClassesForConfigurationUpdates(int batchSize) {
...
+            activeHooks = hooks.get(clazz);
+        } catch (Exception e) {
+            //should actually never occur
+            log.error("Error updating method hook", e);
+            return;
+        }
 
-        deactivateRemovedHooks(clazz, hookConfigs, activeClassHooks);
-        addOrReplaceHooks(clazz, hookConfigs, activeClassHooks);
+        deactivateRemovedHooks(clazz, hookConfigs, activeHooks);
```
ivansenic: seems you overwrote your own fixes in this class here, I am not even sure if this class should be changed at all in this commit?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/hook/MethodHookGenerator.java
```diff
@@ -59,11 +87,93 @@ public void execute(IHookAction.ExecutionContext ctx) {
...
+        ResolvedGenericDataProviderConfig providerConfig = providerCallConfig.getProvider();
+        val injectedProviderClass = dataProviderGenerator.getOrGenerateDataProvider(providerConfig, contextClass);
+
+        val dynamicAssignments = getDynamicInputAssignments(providerCallConfig);
+        val constantAssignments = getConstantInputAssignments(providerCallConfig, contextClass.getClassLoader());
+
+        return BoundDataProvider.bind(dataKey, providerConfig, injectedProviderClass, constantAssignments, dynamicAssignments);
+    }
+
+    private Map<String, Object> getConstantInputAssignments(ResolvedDataProviderCall providerCallConfig, ClassLoader context) {
```
ivansenic: these two methods are not fully clear still

PR #114, opened 2019-02-22
Enabling self-monitoring in demo scenarios

AlexanderWert:


inspectit-oce-demo/agentconfig/customers-service/config.yml
```diff
@@ -1,2 +1,27 @@
 # See https://github.com/inspectIT/inspectit-oce/blob/0.1.M1/inspectit-oce-core/src/main/resources/config/default.yml
-# for all the configuration options.
\ No newline at end of file
+# for all the configuration options.
+inspectit:
+  self-monitoring:
+    enabled: true
+#  instrumentation:
```
JonasKunz: Should these commented out lines stay in here? if yes maybe you should update the descriptions on them.

inspectit-oce-demo/grafana/provisioning/elastic/dashboards/self-monitoring.json
```diff
@@ -0,0 +1,2096 @@
+{
```
JonasKunz: We have added a new component to inspectit with a new working-queue of classes with the following emtric:
inspectit_self_instrumentation_hooking_queue_size
You should use the maximum of this and the instrumentation queue size to compute the progress.

inspectit-oce-demo/prometheus/prometheus.yml
```diff
@@ -1,6 +1,6 @@
 # my global config
 global:
-  scrape_interval:     15s
+  scrape_interval: 1s
```
JonasKunz: We have configured inspectit.metrics.frequency by default with 15s. If you want system metrics to be updated at a higher rate you should configure this to a lower value for each service.

PR #121, opened 2019-02-25
jmh setup

ivansenic:
* before we merge I will make my decision if we should specify the benchmark mode on the test class, or by default always use the default which is throughput.. I am not sure in general what are we more interested in?
*  I updated Gradle to `5.2.1`
* read the `PERFTEST.md` first 
* please check carefully my updates in the non-test classes

### DIFF
--- a/inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/context/InspectitContext.java
+++ b/inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/context/InspectitContext.java
@@ -211,9 +211,7 @@ public class InspectitContext implements IInspectitContext {
      */
     @Override
     public void makeActive() {
-
-        boolean anyDownPropagatedDataOverwritten = dataOverwrites.keySet().stream()
-                .anyMatch(propagation::isPropagatedDownWithinJVM);
+        boolean anyDownPropagatedDataOverwritten = anyDownPropagatedDataOverridden();
 
         //only copy if any down-propagating value has been written
         if (anyDownPropagatedDataOverwritten) {
@@ -236,6 +234,15 @@ public class InspectitContext implements IInspectitContext {
             }
             activePhaseDownPropagationTagContext = tagger.getCurrentTagContext();
         }
+    }
+
+    private boolean anyDownPropagatedDataOverridden() {
+        for (String key : dataOverwrites.keySet()) {
+            if (propagation.isPropagatedDownWithinJVM(key)) {
+                return true;
+            }
+        }
+        return false;
     }
 
     /**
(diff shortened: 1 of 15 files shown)


inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/context/InspectitContext.java
```diff
@@ -211,12 +211,10 @@ public static InspectitContext createFromCurrent(CommonTagsManager commonTagsMan
...
     @Override
     public void makeActive() {
-
-        boolean anyDownPropagatedDataOverwritten = dataOverwrites.keySet().stream()
-                .anyMatch(propagation::isPropagatedDownWithinJVM);
+        boolean anyDownPropagatedDataOverwritten = anyDownPropagated();
 
         //only copy if any down-propagating value has been written
-        if (anyDownPropagatedDataOverwritten) {
+        if (anyDownPropagated()) {
```
JonasKunz: you are not using the boolean variable here which you declared above

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/context/InspectitContext.java
```diff
@@ -238,6 +236,15 @@ public void makeActive() {
         }
     }
 
+    private boolean anyDownPropagated() {
```
JonasKunz: rename maybe to "anyDownPropagatedDataOverridden" ?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/context/InspectitContext.java
```diff
@@ -384,13 +391,11 @@ private void readOverridesFromCurrentTagContext() {
 
     private Stream<Map.Entry<String, Object>> getDataAsStream() {
         return Stream.concat(
-                postEntryPhaseDownPropagatedData.entrySet().stream()
-                        .filter(e -> !dataOverwrites.containsKey(e.getKey())),
-                dataOverwrites.entrySet().stream()
-                        .filter(e -> e.getValue() != null));
+                postEntryPhaseDownPropagatedData.entrySet().stream().filter(e -> !dataOverwrites.containsKey(e.getKey())),
```
JonasKunz: did you intentionally remove the line breaks here?

JonasKunz:
I somehow get a "Error reading benchmark list" after excecuting the JMH task for the first time successfully.
Even a gradle clean does not help.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/context/InspectitContext.java
ivansenic: :+1: great catch..

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/context/InspectitContext.java
ivansenic: done

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/context/InspectitContext.java
ivansenic: hmm seems so, maybe I tried to optimize here and then reverted it.. Look easier to read for `Stream.concat`, should I revert?

JonasKunz: Matter of taste, I'm fine with both

PR #122, opened 2019-02-27
Stability fixes for non initializable classes, improved type matching of scopes.

JonasKunz:


inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationScopeResolver.java
```diff
@@ -72,7 +72,8 @@ private InstrumentationScope resolveScope(InstrumentationScopeSettings scopeSett
             methodMatcher = any();
         }
 
-        return new InstrumentationScope(typeMatcher, methodMatcher);
+        //we ensure that we only match types which contain at least one matched method
+        return new InstrumentationScope(typeMatcher.and(declaresMethod(methodMatcher)), methodMatcher);
```
mariusoe: Can you separate this for readability so that it is like

```

        //we ensure that we only match types which contain at least one matched method
        typeMatcher = typeMatcher.and(declaresMethod(methodMatcher));
        
        return new InstrumentationScope(typeMatcher, methodMatcher);
```

PR #128, opened 2019-03-07
Gave DataProviders access to the method name, class and parameter types.

JonasKunz:


inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/hook/MethodReflectionInformation.java
```diff
@@ -0,0 +1,80 @@
...
+                .ifPresent(reflMethod ->
+                        builder.parameterTypes(Stream.of(reflMethod.getParameterTypes())
+                                .map(c -> new WeakReference<Class<?>>(c))
+                                .collect(Collectors.toList())));
+        return builder.build();
+    }
+
+
+    public Class<?>[] getParameterTypes() {
+        Class<?>[] paramTypesArray = new Class[parameterTypes.size()];
```
mariusoe: `return (Class<?>[]) parameterTypes.toArray();` ;)

inspectit-oce-core/src/main/resources/config/default.yml
```diff
@@ -131,10 +131,17 @@ inspectit:
           sinceNanos: long
         value: "new Double( (System.nanoTime() - sinceNanos) * 1E-6)"
 
+      get_method_fqn:
+        input:
+          methodName: String
+          clazz: Class
+        value: new StringBuilder(clazz.getName()).append('.').append(methodName).toString()
```
mariusoe: Interessting. I just read this article: https://dzone.com/articles/string-concatenation-performacne-improvement-in-ja

Apparently, java will compile simple `+` operations into a stringbuilder, so, you also could use just `clazz.getName() + '.' + methodName` here if you want.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/hook/MethodHookGenerator.java
```diff
@@ -89,52 +87,51 @@ private void addMetricsRecorder(MethodHookConfiguration config, CopyOnWriteArray
...
      * Generates a data provider and binds its arguments.
      *
-     * @param contextClass       the class in which this data provider will be used.
+     * @param method             the method in which this data provider will be used.
      * @param dataKey            the name of the data whose value is defined by executing the given data provider
      * @param providerCallConfig the specification of the call to the data provider
      * @return the executable data provider
      */
-    private BoundDataProvider generateAndBindDataProvider(Class<?> contextClass, String dataKey, DataProviderCallConfig providerCallConfig) {
+    private BoundDataProvider generateAndBindDataProvider(MethodReflectionInformation method, String dataKey, DataProviderCallConfig providerCallConfig) {
```
mariusoe: Can you please name `method` to `methodInformation` in the methods of this class?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/hook/MethodHookGenerator.java
```diff
@@ -89,52 +87,51 @@ private void addMetricsRecorder(MethodHookConfiguration config, CopyOnWriteArray
...
             val providerCallConfig = pair.getRight();
             try {
-                BoundDataProvider call = generateAndBindDataProvider(hookedClass, dataKey, providerCallConfig);
+                BoundDataProvider call = generateAndBindDataProvider(method, dataKey, providerCallConfig);
                 entryActions.add(call);
             } catch (Exception e) {
-                log.error("Failed to build entry data provider {} for data {} on method {} of {}, no value will be assigned",
-                        providerCallConfig.getProvider().getName(), dataKey, signature, hookedClass.getName(), e);
+                log.error("Failed to build entry data provider {} for data {} on method {}.{}, no value will be assigned",
+                        providerCallConfig.getProvider().getName(), dataKey, method.getDeclaringClass().getName(), method.getName(), e);
```
mariusoe: `method.getMethodFQN()`?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/hook/MethodHookGenerator.java
```diff
@@ -89,52 +87,51 @@ private void addMetricsRecorder(MethodHookConfiguration config, CopyOnWriteArray
...
             val providerCallConfig = pair.getRight();
             try {
-                BoundDataProvider call = generateAndBindDataProvider(hookedClass, dataKey, providerCallConfig);
+                BoundDataProvider call = generateAndBindDataProvider(method, dataKey, providerCallConfig);
                 exitActions.add(call);
             } catch (Exception e) {
-                log.error("Failed to build exit data provider {} for data {} on method {} of {}, no value will be assigned",
-                        providerCallConfig.getProvider().getName(), dataKey, signature, hookedClass.getName(), e);
+                log.error("Failed to build exit data provider {} for data {} on method {}.{}, no value will be assigned",
+                        providerCallConfig.getProvider().getName(), dataKey, method.getDeclaringClass().getName(), method.getName(), e);
```
mariusoe: See comment above

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/dataproviders/GenericDataProviderSettings.java
```diff
@@ -21,19 +21,30 @@
 public class GenericDataProviderSettings {
 
     public static final String PACKAGE_REGEX = "[a-zA-Z]\\w*(\\.[a-zA-Z]\\w*)*";
+
     public static final String THIZ_VARIABLE = "thiz";
     public static final String ARGS_VARIABLE = "args";
     public static final String THROWN_VARIABLE = "thrown";
     public static final String RETURN_VALUE_VARIABLE = "returnValue";
     public static final String ARG_VARIABLE_PREFIX = "arg";
     public static final Pattern ARG_VARIABLE_PATTERN = Pattern.compile(ARG_VARIABLE_PREFIX + "(\\d+)");
```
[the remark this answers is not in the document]
JonasKunz: This is stored as compiled pattern, because it is used for regex-matching in the DataProviderGenerator. I think storing a regex as public-static-final makes little sense as you will only use it in compiled form anyway.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/dataproviders/GenericDataProviderSettings.java
```diff
@@ -21,19 +21,30 @@
 public class GenericDataProviderSettings {
 
     public static final String PACKAGE_REGEX = "[a-zA-Z]\\w*(\\.[a-zA-Z]\\w*)*";
+
     public static final String THIZ_VARIABLE = "thiz";
```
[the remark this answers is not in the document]
JonasKunz: I agree, very good idea!

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/hook/MethodReflectionInformation.java
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * @return The class on which this method hook is applied, never null as long as the class has not been garbage collected.
+     */
+    public Class<?> getDeclaringClass() {
+        return declaringClass.get();
+    }
+
+
+    public String getMethodFQN() {
+        return getDeclaringClass().getName() + "." + getName();
```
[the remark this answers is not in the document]
JonasKunz: This method is only called from within method hooks, meaning that the class can never be gced.
But let's be safe ;)

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/hook/MethodReflectionInformation.java
JonasKunz: parameterTypes is an array of weak reference, which we want to resolve prior to returning the array.
And we don't want to use streams for performance reasons here, therefore this solution.

inspectit-oce-core/src/main/resources/config/default.yml
JonasKunz: Well the java compiler will, however javassist has its own compiler implementation. I'm not sure if they also implement this optimization, therefore I will leave it as Stringbuilder for now

PR #131, opened 2019-03-11
Closes #116 - Added classloader delegation

### LINKED ISSUE #116: Add special instrumentation for OSGi-style classloaders
Our instrumentation assumes that our bootstrap classes are accessible from every application classloader.
This is not always the case, e.g. in OSGi. In this cases our classes would most likely result in the application terminating with NoClassDefErrors.

We can prevent this by checkling if our bootstrap classes are reachable from the target classloader before performing an instrumentation. In case we discover that our bootstrap classes are not reachable we could add a classlaoder delegation similar to the old inspectit.

JonasKunz:
After our initial discussion I decided to change the instrumentation in the following way:
I extracted the logic of handling the queue of which classes need to be checked from the InstrumentationManager and the HookManager to a single queue managed by the new class "InstumentationTriggerer" (name is discussable ... :D ).
This class is then also responsible for ensuring that the class loader delegation has been applied before we try to do any instrumentation / hooking.


inspectit-oce-demo/grafana/provisioning/influxdb/dashboards/self-monitoring.json
```diff
@@ -16,7 +16,7 @@
   "gnetId": null,
```
ivansenic: what about the elastic dashboard?

JonasKunz:
I now tested it with the following application:
http://felix.apache.org/documentation/subprojects/apache-felix-ipojo/apache-felix-ipojo-gettingstarted/ipojo-hello-word-maven-based-tutorial.html
``` 
java -jar -javaagent:/home/jku/inspectit-oce/inspectit-oce-agent/build/inspectit-oce-agent-SNAPSHOT.jar 
-Dinspectit.logging.debug=true 
-Dinspectit.instrumentation.scopes.myscope.type.name=ipojo.example -Dinspectit.instrumentation.scopes.myscope.type.matcher-mode=STARTS_WITH 
-Dinspectit.instrumentation.rules.record_method_duration.scopes.myscope=true 
-Dinspectit.self-monitoring.enabled=true 
bin/felix.jar
```
when disablingthe CLD, the app crashes with an NoClassDefFound error, when enabling it the resposne times metrics are recorded correctly.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/special/ClassLoaderDelegation.java
```diff
@@ -1,165 +1,103 @@
...
-
-        Boolean isBootstrapAccessible = bootstrapClassesAccessible.getIfPresent(classLoader);
-        if (isBootstrapAccessible == null) {
-            return makeBootstrapClassesAvailableSynchronized(classLoader);
-        } else {
-            return isBootstrapAccessible;
-        }
+    @Override
+    public boolean requiresInstrumentationChange(Class<?> clazz, InstrumentationConfiguration first, InstrumentationConfiguration second) {
+        return false;
```
ivansenic: This remaining false is still correct? We will not react to the configuration updates?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationConfigurationResolver.java
```diff
@@ -111,6 +111,9 @@ public ClassInstrumentationConfiguration getClassInstrumentationConfiguration(Cl
      */
     public Map<MethodDescription, MethodHookConfiguration> getHookConfigurations(Class<?> clazz) {
         val config = currentConfig;
+        if (isIgnoredClass(clazz, config)) {
```
ivansenic: is this anyhow related to this ticket?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/special/ClassLoaderDelegation.java
JonasKunz: Yes, we do react via shouldInstrument, this method only tests if changes are necessary while the sensor remained active.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationConfigurationResolver.java
JonasKunz: It's a bug I discovered. We excluded ignored classes from instrumentation but not from hooking.
I saw during testing in the debug output that hooks were built for our own injected classes... :D

PR #133, opened 2019-03-12
Added conditions and before option to customize data provider order

JonasKunz:


inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/DataProviderCallScheduler.java
```diff
@@ -0,0 +1,158 @@
...
+                .collect(Collectors.toList());
+    }
+
+    private void collectDependencies(String dataKey, DataProviderCallConfig providerCall, Map<String, Set<String>> dependencyGraph) {
+        DataProviderCallSettings callSettings = providerCall.getCallSettings();
+
+        Set<String> afterDependencies = new HashSet<>();
+        //this data provider uses the following data keys as input arguments, therefore depends on them
+        val dataInputs = callSettings.getDataInput().values();
+        afterDependencies.addAll(dataInputs);
```
mariusoe: For me, `afterDependencies` and `beforeDependencies` was quite confusing at the beginning because `beforeDependencies` are dps (data providers) which are called after the actual one, right?

Can we rename those to (I have no good idea), for example:
`afterDependencies` => `dependencies` or `preconditions`
`beforeDependencies` => `followupDataProviders`, `followupDependencies` (I dont know if this makes sense :D )

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/MethodHookConfigurationResolver.java
```diff
@@ -14,16 +12,21 @@
...
-import java.util.stream.Collectors;
 
 @Component
 public class MethodHookConfigurationResolver {
 
     @Autowired
     DataProviderGenerator providerGenerator;
 
+    @Autowired
+    DataProviderCallScheduler scheduler;
```
mariusoe: I really like that this is separated!

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/hook/ConditionalHookAction.java
```diff
@@ -0,0 +1,29 @@
+package rocks.inspectit.oce.core.instrumentation.hook;
+
+import lombok.Value;
+
+@Value
+public class ConditionalHookAction implements IHookAction {
+
+
```
mariusoe: Remove second new line

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/hook/ConditionalHookAction.java
```diff
@@ -0,0 +1,29 @@
...
+    @FunctionalInterface
+    public interface Condition {
+        boolean evaluate(IHookAction.ExecutionContext context);
+    }
+
+    private Condition condition;
+
+    private IHookAction action;
```
mariusoe: I dont know how lombok is generating the `@Value` code, but should we add `@NonNull` to the fields?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/hook/MethodHookGenerator.java
```diff
@@ -122,14 +123,45 @@ private void addDataProviderCalls(MethodReflectionInformation method, MethodHook
...
-        return BoundDataProvider.bind(dataKey, providerConfig, injectedProviderClass, constantAssignments, dynamicAssignments);
+        IHookAction providerCall = BoundDataProvider.bind(dataKey, providerConfig, injectedProviderClass, constantAssignments, dynamicAssignments);
+
+        return addConditionsToProviderCall(callSettings, providerCall);
+    }
+
+    private IHookAction addConditionsToProviderCall(DataProviderCallSettings callSettings, IHookAction providerCall) {
+        if (!StringUtils.isEmpty(callSettings.getOnlyIfTrue())) {
+            String conditionDataKey = callSettings.getOnlyIfTrue();
+            providerCall = new ConditionalHookAction((e) -> {
```
mariusoe: Can you change var name `e` to `ctx`

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/hook/MethodHookGenerator.java
```diff
@@ -122,14 +123,45 @@ private void addDataProviderCalls(MethodReflectionInformation method, MethodHook
...
+                Object val = e.getInspectitContext().getData(conditionDataKey);
+                return val != null && (Boolean) val;
```
mariusoe: If the data key is referencing something which is not a boolean it will throw an exception. Is this by purpose of should we use `Boolean.valueOf(..)`? This will not throw an exception but has less transparency why a value might be false all the time?.

inspectit-oce-agent/src/system-test/resources/config/EntryPointDetectionTest.yml
```diff
@@ -0,0 +1,46 @@
...
+        scopes:
+          EntryPointDetectionTest-methods: true
+        entry:
+          method_name: {provider: get_method_name}
+          entry_point_found: {provider: assign, constant-input: {val: true}}
+          is_entry_point:
+            only-if-null: entry_point_found
+            before:
+              entry_point_found : true
+            provider: assign
```
mariusoe: Would it make sense to move the `assign` dp to the actual agent, thus, it is not only for testing existing?

inspectit-oce-agent/src/system-test/java/rocks/inspectit/oce/instrumentation/EntryPointDetectionTest.java
```diff
@@ -0,0 +1,41 @@
...
+                .isInstanceOfSatisfying(AggregationData.CountData.class, data -> assertThat(data.getCount()).isEqualTo(1));
+        assertThat(TestUtils.getDataForView("entrypoint/invocations", Maps.newHashMap("method_name", "methodB")))
+                .isNull();
+
+        //here methodB should be recognized as entry point
+        methodB();
+
+        TestUtils.waitForOpenCensusQueueToBeProcessed();
+        assertThat(TestUtils.getDataForView("entrypoint/invocations", Maps.newHashMap("method_name", "methodA")))
+                .isInstanceOfSatisfying(AggregationData.CountData.class, data -> assertThat(data.getCount()).isEqualTo(1));
```
mariusoe: This is still one because we are not "resetting" the metric, right? Just resolve if this is true, otherwise ping me :)

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/hook/MethodHookGenerator.java
```diff
@@ -122,14 +123,45 @@ private void addDataProviderCalls(MethodReflectionInformation method, MethodHook
...
         val dynamicAssignments = getDynamicInputAssignments(methodInfo, providerCallConfig);
         val constantAssignments = getConstantInputAssignments(methodInfo, providerCallConfig);
 
-        return BoundDataProvider.bind(dataKey, providerConfig, injectedProviderClass, constantAssignments, dynamicAssignments);
+        IHookAction providerCall = BoundDataProvider.bind(dataKey, providerConfig, injectedProviderClass, constantAssignments, dynamicAssignments);
+
+        return addConditionsToProviderCall(callSettings, providerCall);
+    }
+
+    private IHookAction addConditionsToProviderCall(DataProviderCallSettings callSettings, IHookAction providerCall) {
```
mariusoe: Please add a short doc

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/DataProviderCallScheduler.java
```diff
@@ -0,0 +1,158 @@
...
+ * The approach is that first a dependency graph is generated.
+ * In this dependency graph the nodes are the data-keys and the data provider calls writing them,
+ * the edges indicate a dependency between data keys and the corresponding data provider calls.
+ * The edges are direct and and edge from data_A to data_B reads as
+ * "The provider writing data_B must be executed before the provider of data_A"
+ */
+@Component
+public class DataProviderCallScheduler {
+
+    public List<Pair<String, DataProviderCallConfig>> orderDataProviderCalls(Map<String, DataProviderCallConfig> calls) throws CyclicDataDependencyException {
```
mariusoe: Can you add some docs for this please

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/DataProviderCallScheduler.java
```diff
@@ -0,0 +1,158 @@
...
+                    int idx = stackTraceList.indexOf(dependency);
+                    throw new CyclicDataDependencyException(stackTraceList.subList(idx, stackTraceList.size()));
+                }
+            }
+        }
+        result.push(current);
+        stackTrace.removeTop();
+    }
+
+    static class StringStack {
```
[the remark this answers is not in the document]
JonasKunz: ArrayDeque doesn't have a O(1) contains as stated in the documentation:
"Most ArrayDeque operations run in amortized constant time. Exceptions include remove, removeFirstOccurrence, removeLastOccurrence, contains, iterator.remove(), and the bulk operations, all of which run in linear time."
This was the actual reason I implemented this combination class.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/DataProviderCallScheduler.java
```diff
@@ -0,0 +1,158 @@
...
+/**
+ * This class encapsulates the logic how data providers are ordered.
+ * The approach is that first a dependency graph is generated.
+ * In this dependency graph the nodes are the data-keys and the data provider calls writing them,
+ * the edges indicate a dependency between data keys and the corresponding data provider calls.
+ * The edges are direct and and edge from data_A to data_B reads as
+ * "The provider writing data_B must be executed before the provider of data_A"
+ */
+@Component
+public class DataProviderCallScheduler {
```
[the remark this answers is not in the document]
JonasKunz: Your proposed name is fine for me. I though on compiler level where you have a nexecution scheduler which reorders your operations :D

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/hook/MethodHookGenerator.java
JonasKunz: Boolean.valueOf only accepts "Strings", whereas the value could be any type off Object if misused.
I on purpose only allowed "Booleans".
When this is not a boolean, an exception will be raised. This exception is caught by the MethodHook, which prints out an error to the log and then disabled the erronous data provider.

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/DataProviderCallScheduler.java
JonasKunz: How about the following names:
afterDependencies: executedAfter
beforeDependencies: executedBefore
?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/hook/ConditionalHookAction.java
JonasKunz: `@Value` generates an all args constructor and therefore forces you to specify the values.

inspectit-oce-agent/src/system-test/resources/config/EntryPointDetectionTest.yml
JonasKunz: Then we should discuss the name, if assign is really the best choice. Other candidates are:
- copy
- set
- something else?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/DataProviderCallScheduler.java
mariusoe: Passt!

PR #136, opened 2019-03-13
Closes #130 - Add option to exclude lambdas from instrumentation

### LINKED ISSUE #130: Add option to ignore lambda expressions
As we run into problems again and again when trying to instrument certain Lambda expressions, we should add an option to dsiable the instrumentation of Lambdas. This option should be enabled by default.

mariusoe:


inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/instrumentation/config/InstrumentationScopeResolver.java
```diff
@@ -73,7 +73,11 @@ private InstrumentationScope resolveScope(InstrumentationScopeSettings scopeSett
         }
 
         //we ensure that we only match types which contain at least one matched method
-        typeMatcher = typeMatcher.and(declaresMethod(methodMatcher));
+        if (typeMatcher.equals(any())) {
```
JonasKunz: What's the intention behind this change?

inspectit-oce-core/src/main/java/rocks/inspectit/oce/core/config/model/instrumentation/scope/AdvancedScopeSettings.java
```diff
@@ -24,4 +24,8 @@
      */
     private boolean disableSafetyMechanisms = false;
 
+    /**
+     * Defines whether lambda classes should be excluded from the declared scope.
+     */
+    private boolean excludeLambdas = true;
```
JonasKunz: I would rather put in on the instrumentation level, next to "ignore-packages" I would add an "ignore-lambdas",
Otherwise it would be impossible to exclude lambdas being instrumented with special sensors.
Then you can also jsut keep your matchers as they are and add a corresponding check to InstrumentationConfigurationResolver.isIgnoredClass()

PR #138, opened 2019-03-18
Closes #134 - Removed elastic demo scenario

### LINKED ISSUE #134: Remove the elastic demo scenario
The elastic scenario currently is not quiet working and also doesn't support certain essential features, like computing the average from a sum and a count metric. For this reason we should remove it.

AlexanderWert:


mariusoe:
Please change the name to a nice an expressive name because it will be in the changelog!

PR #148, opened 2019-03-22
Renamed project and related files from OCE to Ocelot

mariusoe:
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-oce/148)


JonasKunz:
I did not check all the deltas but I checked out the PR locally and built it and verified that the demo works :)

PR #166, opened 2019-04-24
Removed the Flux data source from the influxDB demo

JonasKunz:
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/166)


mariusoe:
Please also remove the directory `grafana\provisioning\influxdb\flux-dashboards`. We don't need it anymore when the FLUX ds is gone.

PR #181, opened 2019-05-02
Ensured that tests wait five seconds for instrumentations to become active

JonasKunz:
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/181)

### DIFF
--- a/inspectit-ocelot-agent/src/system-test/java/rocks/inspectit/ocelot/utils/TestUtils.java
+++ b/inspectit-ocelot-agent/src/system-test/java/rocks/inspectit/ocelot/utils/TestUtils.java
@@ -16,6 +16,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


inspectit-ocelot-agent/src/system-test/java/rocks/inspectit/ocelot/utils/TestUtils.java
```diff
@@ -27,6 +28,29 @@
 
     private static Cache<Class<?>, Object> activeInstrumentations = null;
 
+    public static ConcurrentHashMap<Class<?>, Long> instrumentationTimeStamp = new ConcurrentHashMap<>();
+
+    static {
+        Thread poller = new Thread(() -> {
+
```
mariusoe: Can you remove this empty line please :)

inspectit-ocelot-agent/src/system-test/java/rocks/inspectit/ocelot/utils/TestUtils.java
```diff
@@ -76,20 +100,31 @@ public static void waitForClassInstrumentations(List<Class> clazzes, int duratio
             await().atMost(duration, timeUnit).ignoreExceptions().untilAsserted(() -> {
                 for (Class clazz : clazzes) {
                     sink = Class.forName(clazz.getName(), true, clazz.getClassLoader());
-                    Object clazzInstrumentation = getInstrumentationCache().getIfPresent(clazz);
-                    assertThat(clazzInstrumentation).isNotNull();
+                    Long timeStamp = instrumentationTimeStamp.get(clazz);
```
mariusoe: I think this has to be reverted and we have to check the actual cache here. Because otherwise, assuming a class' instrumentation has been reverted, we would still consider it as instrumented because it exists in our local cache.
Maybe it's quite unlikly but I think there is no benefit checking the local cache (the one in this class)

PR #345, opened 2019-08-13
Closes #276: Renew authetication token when expired

### LINKED ISSUE #276: Renew authetication token when expired
If the authentication token of the currently logged in user expires it should be renewd. This has to be done using the current token, thus, the UI does not need the user credentials.

tangerstein:
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/345)


tangerstein:
@mariusoe  I've implemented your comments :)

tangerstein:
@mariusoe  If updated the pull request according to our latest discussion

PR #412, opened 2019-09-16
Added memory limit to gradle

JonasKunz:
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/412)


JonasKunz:
Did not have any effect...

PR #422, opened 2019-09-24
Closes #350 - downloading config files/ changed agent mappings layout

### LINKED ISSUE #350: Functionallity for exporting the configuration of specific agents using the WebUI
The web UI should be extended by the functionallity to export the actual configuration which would be provided for a specific agent.

It should be able to download the configuration as a YML file - including the content (configuration) which would be received by the agent.

In support cases, this can be helpful because users can export the configuration which is received by an agent to simplify the process of finding errors and problems.

ArianeKraus:
Since I can only attach one Issue I went for the existing one. However this PR changes page layout for displaying/editing and adding agent mappings

---
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/422)


ArianeKraus:
___
*[components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/AgentMappingsView.js, line 18 at r2](https://reviewable.io/reviews/inspectit/inspectit-ocelot/422#-Lrh0wbi5g2YiH8zs9IU:-Lrj61gTErz-HwOXOHaK:bxkb1v0) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/b96e1ffb0397a114d11b3edd8953e2f9f86a9f72/components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/AgentMappingsView.js#L18)):*

Not needed anymore?
Also remove the corresponding callbacks and the refresh dialog below
</blockquote></details>

would a refresh button still be useful (just without the dialog) or can it be omitted? 
I could imagine it useful in case someone else is able to change mappings at the same time... 
Without a dialog a 'last edited xx-xx' timestamp would probably be reasonable

ArianeKraus:
___
*[components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/MappingsTable.js, line 69 at r2](https://reviewable.io/reviews/inspectit/inspectit-ocelot/422#-Lrh6cVX-FMVCeYrSEDK:-LrjC0OZ7eU8_j_HcS4N:b-dczlaz) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/b96e1ffb0397a114d11b3edd8953e2f9f86a9f72/components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/MappingsTable.js#L69)):*

can you explain me why you need to append this to the body?
</blockquote></details>



not appending it at all has some strange behavior of the component -> but since i added the delete dialog to the mappingTable class i had to add a div surrounding dataTable anyway. Appending it to that div seems to work as well, if that's a better solution?

ArianeKraus:
___
*[components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/MappingsTable.js, line 135 at r2](https://reviewable.io/reviews/inspectit/inspectit-ocelot/422#-Lrh9Jhu5uQVBctVusu9:-LrmWwQC5egX97Na2ABb:b-49rmii) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/b96e1ffb0397a114d11b3edd8953e2f9f86a9f72/components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/MappingsTable.js#L135)):*

What is `element.head`?
</blockquote></details>

came from an older version, forgot to delete it

ArianeKraus:
___
*[components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/dialogs/DownloadDialog.js, line 23 at r2](https://reviewable.io/reviews/inspectit/inspectit-ocelot/422#-LrhBLZC90xRhXOBqPEY:-LrmrQNK-dfIbeBOPToH:b-h9002g) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/b96e1ffb0397a114d11b3edd8953e2f9f86a9f72/components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/dialogs/DownloadDialog.js#L23)):*

I think this function can be simplified a lot using something liek this:
```
let newAttributes = cloneDeep(this.state.attributes);
delete newAttributes[oldKey];
newAttributes[newKey] =  this.state.attributes[oldKey] || "";
```
</blockquote></details>

that way the attributes order gets mixed so new attributes might not show up last and changing an attribute key in the middle might result in the attribute switching place...

ArianeKraus:
___
*[components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/dialogs/EditDialog.js, line 19 at r2](https://reviewable.io/reviews/inspectit/inspectit-ocelot/422#-LrhCdNr7mpDmfJahj3u:-LrnKY65D9BIsY0TGVQ2:bfnrdzs) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/b96e1ffb0397a114d11b3edd8953e2f9f86a9f72/components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/dialogs/EditDialog.js#L19)):*

After seeing it I only think a `FieldSet` is required around the sources part, the attributes and the name input don't need it.
You could however add a label in front of the name field.
</blockquote></details>



Personally I like it more when there is a Fieldset around attribute as well

ArianeKraus:
___
*[components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/dialogs/DownloadDialog.js, line 23 at r2](https://reviewable.io/reviews/inspectit/inspectit-ocelot/422#-LrhBLZC90xRhXOBqPEY:-LryCvzn7E99uizgIaq4:bieh5co) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/b96e1ffb0397a114d11b3edd8953e2f9f86a9f72/components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/dialogs/DownloadDialog.js#L23)):*
<details><summary><i>Previously, ArianeKraus (Ariane Kraus) wrote…</i></summary><blockquote>

that way the attributes order gets mixed so new attributes might not show up last and changing an attribute key in the middle might result in the attribute switching place...
</blockquote></details>

don't mind me. No matter how it's still strange :dagger:

ArianeKraus:
___
*[components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/editComponents/EditAttributes.js, line 9 at r2](https://reviewable.io/reviews/inspectit/inspectit-ocelot/422#-LrgzFI63vqdWL6_woK4:-LryE7RI2i7t9jD-x_zF:b-3ati76) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/b96e1ffb0397a114d11b3edd8953e2f9f86a9f72/components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/editComponents/EditAttributes.js#L9)):*

There is a bug in this component causing the browser to hang:
Add an attribute by clicking in the key field and type something. Then click into the value field -> the browser hangs.
</blockquote></details>

changed the function anyway -> hopefully works now

ArianeKraus:
___
*[components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/editComponents/EditAttributes.js, line 9 at r2](https://reviewable.io/reviews/inspectit/inspectit-ocelot/422#-LrhAVws-oX313MbatPB:-LryH0U70a4EPARKj3zR:balcka2) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/b96e1ffb0397a114d11b3edd8953e2f9f86a9f72/components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/editComponents/EditAttributes.js#L9)):*

I would make the allowed inputs more flexible:
You can allow either attribute or value keys to be empty.

The validation that such values are not allowed shuld then happen when the parent dialog is closed.

This removes the need to insert "key" and "value" automaticalyl as you do it right now.
</blockquote></details>

done - however the API it allows storing empty keys/values so i could simply allow it as well

ArianeKraus:
___
*[components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/dialogs/EditDialog.js, line 135 at r6](https://reviewable.io/reviews/inspectit/inspectit-ocelot/422#-LsRZBdAEjULNagNAtY2:-LsWVbN_D2WUFez2WyLS:b9brdrf) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/8b890be949116df0c214be6b5b7186bc1e5f7f99/components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/dialogs/EditDialog.js#L135)):*

You should show a warning and prevent a save / the dialog from closing
</blockquote></details>

within the download dialog as well then?

ArianeKraus:
___
*[components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/editComponents/utils.js, line 21 at r6](https://reviewable.io/reviews/inspectit/inspectit-ocelot/422#-LsS8uns5h3NGns-jo3m:-LsqqazWES1KuZ1S0fyL:b-5xz08g) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/8b890be949116df0c214be6b5b7186bc1e5f7f99/components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/editComponents/utils.js#L21)):*

This looks buggy to me:
Imagine having two children, where the target node is found in the first one.
`res` would get overwritten two null here because you don't retur nas soon as you have found a result.
</blockquote></details>

seems like returning the function directly is fine as well

ArianeKraus:
___
*[components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/editComponents/SourceToolbar.js, line 57 at r6](https://reviewable.io/reviews/inspectit/inspectit-ocelot/422#-LsViGBuEPFAg-zgv5U5:-Lszu7uHETctBlIwv219:b-d5jgr8) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/8b890be949116df0c214be6b5b7186bc1e5f7f99/components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/editComponents/SourceToolbar.js#L57)):*

Again no idea what's happening here
</blockquote></details>

changed the function && added comments

ArianeKraus:
___
*[components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/editComponents/SourceTree.js, line 7 at r6](https://reviewable.io/reviews/inspectit/inspectit-ocelot/422#-LsVjWZq2Ggf86AbSnaR:-Lt--Ciz7NigTWKws9bw:b-5ng1mt) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/8b890be949116df0c214be6b5b7186bc1e5f7f99/components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/editComponents/SourceTree.js#L7)):*

Add comments what the functions do. The cleanup logic for the tree is very much spread around multiple components...
I think SourceTree and SOurceTable should be pretty dumb.
The EditSources component should solely be responsible for doing cleanup stuff on selection change.
</blockquote></details>

--> added Comments

ArianeKraus:
___
*[components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/editComponents/utils.js, line 76 at r6](https://reviewable.io/reviews/inspectit/inspectit-ocelot/422#-LsVd8-b0PG47nXUJnfs:-Lt--vvj7G6hEc4Vf6MG:b-dhr2dm) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/8b890be949116df0c214be6b5b7186bc1e5f7f99/components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/editComponents/utils.js#L76)):*

I think the more correct name would be "getParentKeys".
Also you can implement this more easily using a recursve function which invokes `getParentKey`.
</blockquote></details>

changed the function... not recursive tough

ArianeKraus:
___
*[components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/editComponents/utils.js, line 36 at r6](https://reviewable.io/reviews/inspectit/inspectit-ocelot/422#-LsSB1R59Fjz67sezfr1:-Lt-0z5p9sZlVBBbgj9R:bs142tb) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/8b890be949116df0c214be6b5b7186bc1e5f7f99/components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/editComponents/utils.js#L36)):*

This function is really hard to get, can't you simplify this a little? There's so much happening here
</blockquote></details>

commented function

ArianeKraus:
___
*[components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/editComponents/utils.js, line 94 at r6](https://reviewable.io/reviews/inspectit/inspectit-ocelot/422#-LsVdh1z3O3NwdbKJy3z:-Lt-11n70euvppQ5DALT:bdagait) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/8b890be949116df0c214be6b5b7186bc1e5f7f99/components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/editComponents/utils.js#L94)):*

This is a bit hacky, isn't it?
Why not just add a boolean `isFile` input to the `addNode` function?
</blockquote></details>

omitted

ArianeKraus:
___
*[components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/editComponents/utils.js, line 101 at r6](https://reviewable.io/reviews/inspectit/inspectit-ocelot/422#-LsVe2RgCtOh9zTi6KGP:-Lt-18-w9n-rj_YRjwPf:bvnqq01) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/8b890be949116df0c214be6b5b7186bc1e5f7f99/components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/editComponents/utils.js#L101)):*

I have no idea what this function does
</blockquote></details>

--> completely changed

ArianeKraus:
___
*[components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/editComponents/utils.js, line 109 at r6](https://reviewable.io/reviews/inspectit/inspectit-ocelot/422#-LsVeX22ALLUCdBvgOUj:-Lt-1Jq2B3eoDCGwZvaQ:br90brm) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/8b890be949116df0c214be6b5b7186bc1e5f7f99/components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/editComponents/utils.js#L109)):*

This also looks very complex and without comments is hardly understandable.
Also it looks ike there are many special cases, can't you simplify this a little?
</blockquote></details>

function is gone

ArianeKraus:
___
*[components/inspectit-ocelot-configurationserver-ui/src/redux/ducks/agent-config/actions.js, line 33 at r6](https://reviewable.io/reviews/inspectit/inspectit-ocelot/422#-LsVmoQA7Z8yr1Wvla3_:-Lt-1OcI8ACdaH8jg-oZ:bv2xijf) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/8b890be949116df0c214be6b5b7186bc1e5f7f99/components/inspectit-ocelot-configurationserver-ui/src/redux/ducks/agent-config/actions.js#L33)):*

No Ui stuff should ever happen directly within actions!

In fact this action does not really modify the state, so the download can happen directly in your UI component.

You can create a `DownloadConfigurationButton` which manages the hidden link to the blob and performs the download.
</blockquote></details>

created DownloadLink instead of a Button

ArianeKraus:
___
*[components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/dialogs/EditDialog.js, line 136 at r5](https://reviewable.io/reviews/inspectit/inspectit-ocelot/422#-LsLZLs1A8IXkVmYMoNg:-LsWIhVFBU32o2Ty8y5J:b-896fix) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/0549b693a2599cf3b5e81c5f6af1507649882980/components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/dialogs/EditDialog.js#L136)):*

I think you forgot about this one :D
</blockquote></details>

Done.

ArianeKraus:
___
*[components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/MappingsTable.js, line 137 at r7](https://reviewable.io/reviews/inspectit/inspectit-ocelot/422#-Lt94BImBqR-BYBRaaK2:-Lt9cCbW16CtWBBm0Wir:b-h85js1) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/85b144f7732601e9bebdef6f4d456f30f5b0af82/components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/MappingsTable.js#L137)):*

Shouldn't you also use a search-string here like for the attributes because yo uswitched over to the global-filter method?
</blockquote></details>

DataTable is able to read out the source correctly, since it's value is a string --> a search string for sources would look exactly like the original source string
It's only the attribute object which is not readable by DataTable

ArianeKraus:
___
*[components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/MappingsTable.js, line 106 at r7](https://reviewable.io/reviews/inspectit/inspectit-ocelot/422#-Lt93aDwDvykzyPJIXZa:-Lt9d2lBEWIN86o73FVh:b-ylgq3a) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/85b144f7732601e9bebdef6f4d456f30f5b0af82/components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/MappingsTable.js#L106)):*

This is unnecessary, isn't it?
</blockquote></details>

needed within render :)

ArianeKraus:
___
*[components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/MappingsTable.js, line 165 at r7](https://reviewable.io/reviews/inspectit/inspectit-ocelot/422#-Lt94_dZ7-D_BH1IgIsF:-Lt9gfcj8T6eB5teEeak:b-h7ie1s) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/85b144f7732601e9bebdef6f4d456f30f5b0af82/components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/MappingsTable.js#L165)):*

How about moving this component to the Button-Cell component?
It doesn't hurt to have one link per `ButtonCell`.
</blockquote></details>

tried it out (moved ref into buttonCell as well) but it actually resulted in an undefined error in some cases...

ArianeKraus:
___
*[components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/dialogs/EditDialog.js, line 123 at r7](https://reviewable.io/reviews/inspectit/inspectit-ocelot/422#-Lt9EFo9-9jEQfboAStz:-LtA04qUBnFGFWTmOICK:bb7pcsz) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/85b144f7732601e9bebdef6f4d456f30f5b0af82/components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/dialogs/EditDialog.js#L123)):*

shouldn't it be sufficient to just compare the names here?
</blockquote></details>

true

ArianeKraus:
___
*[components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/dialogs/DownloadDialog.js, line 64 at r11](https://reviewable.io/reviews/inspectit/inspectit-ocelot/422#-LvotPga5Vhl0hdnifn-:-Lvp2-3qCHxc0VnIRQ8X:b32viz) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/c8cf4a4f17d47ac75710e6361e6220f765470eab/components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/dialogs/DownloadDialog.js#L64)):*

Again, writing this is bad (even though it apprantely works ... :D )
</blockquote></details>

:P

ArianeKraus:
___
*[components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/dialogs/EditDialog.js, line 152 at r11](https://reviewable.io/reviews/inspectit/inspectit-ocelot/422#-LvotavZ2vSHYjbXckHU:-Lvp2uZN4HzLJ6WQOyWy:by95fpw) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/c8cf4a4f17d47ac75710e6361e6220f765470eab/components/inspectit-ocelot-configurationserver-ui/src/components/views/mappings/dialogs/EditDialog.js#L152)):*

Why did you add this condition?
Maybe you can move the findIndex calculation outside of the if to make it more readable
</blockquote></details>

Before, a new added mapping with a name which already existed would have overwritten the old mapping --> so it gives out an error now instead.

PR #458, opened 2019-11-02
Closes #406 - Prevent log spam if the HTTP config is unavailable

### LINKED ISSUE #406: Prevent log spam if the HTTP config is unavailable
At the current state, the agent prints an error message for every unsuccessful connection attempt to the configured http config URL.
We should add a "failure-counter" and print only on 1,2,4,8... attempt the error message.
In addition we should print a success message on INFO-level for the first successful fetch after a failure.

rexino:
Added log counter and prints only every 1,2,4,8.. (i understand that as every power of two). After successful connect it prints INFO log with success message and number of previous failed attempts.

---
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/458)


rexino:
Thank you very much for your review :) 
Hope it is good this time.

PR #503, opened 2019-12-21
Fix lint problems of the demos' Dockerfiles

mc-chaos:
Updates to Docker-compose files, to remove warnings from yamllint.
See https://github.com/adrienverge/yamllint

---
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/503)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 !gradle/wrapper/gradle-wrapper.jar
 *.log
 *.zip
+*.swp
 
 ### STS ###
 .apt_generated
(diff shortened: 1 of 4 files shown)


mariusoe:
Hello @mc-chaos 
thanks for you contribution and fixing the lint issues of the demo's Dockerfiles.

I just have one suggestion: can you remove the `# yamllint disable rule:line-length` out of the files?
Because I think we should not add this to the files because it could be configured to an appropriate length and due to the fact, that we currenlty don't use `yamllint` in the project. As there are many linters for YAML files availalbe, this should not included in the file, otherwise we nee a lot of these rule-exclutions if we want to support all linters.

mc-chaos:
Hi @mariusoe , 
its me, sascha. I have removed the yamllint commandline.
Regards, Sascha

mariusoe:
Hi @mc-chaos 
ah ok, that's you :)

Thanks for the contribution!

PR #505, opened 2020-01-08
Closes #478 - Make the file-based-config scan subdirectories

### LINKED ISSUE #478: Make the file-basd configuration also scan subdirectories
Currently the file-based configuration does not scan subdirectories for configuration files, which can counterintuitive.
The decision was made because at that time we used the Java WatchService for listerning for changes, which does not support nested directories. We should remove it and entirely rely on the polling mechanism.

ArianeKraus:
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/505)


mariusoe:
In my oppinion, we can remove the watch service and only use and maintain the polling approach - thus we don't have to maintain multiple approaches.

Issue #334, opened 2019-08-06
Unified error responses from configuration server


mariusoe:
Currently, the configuration server returns different structures when errors occure.
The server should provide a unified format for error objects in order to ensure a consistend error handling.

PR #520, opened 2020-01-16
Closes #519 - Add ability to create user via config UI

### LINKED ISSUE #519: Add Users via Config UI
Adding the ability to create new users via the config-servers UI
~ Part of the split PR (#359)

ArianeKraus:
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/520)

### DIFF
--- a/components/inspectit-ocelot-configurationserver-ui/src/components/views/settings/dialogs/CreateDialog.js
+++ b/components/inspectit-ocelot-configurationserver-ui/src/components/views/settings/dialogs/CreateDialog.js
@@ -1,0 +1,124 @@
+import React from 'react';
+import { connect } from 'react-redux';
+
+import { Dialog } from 'primereact/dialog';
+import { Button } from 'primereact/button';
+import { InputText } from 'primereact/inputtext';
+import { Message } from 'primereact/message';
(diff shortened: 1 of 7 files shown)


ArianeKraus:
___
*[components/inspectit-ocelot-configurationserver-ui/src/lib/axios-api.js, line 49 at r1](https://reviewable.io/reviews/inspectit/inspectit-ocelot/520#-LymjcCo-f3qsBbxcFqf:-Lymr08B4QghDmwqm7Xm:b-1b923w) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/ba5257db7a6fa21fd9a01aabcc7942d7eb15f9f5/components/inspectit-ocelot-configurationserver-ui/src/lib/axios-api.js#L49)):*

Can you explain me this change?
</blockquote></details>

When creating a user fails, the config server returns a meaningful message (no username/ no password/ user already exists). (It never does, except for user/account controller) That change is used to access and display the message to the user. Otherwise only a 400 status error will be shown.

That change is kind of connected to this issue: https://github.com/inspectIT/inspectit-ocelot/issues/334 .

PR #522, opened 2020-01-17
Closes #517 - Create separate files for config server logs

### LINKED ISSUE #517: Create separate log files for the configuration server's access and audit log
Currently, the configuration server put all its logs into a single file which makes it quite difficult to look, e.g. for audit or access logs.
In order to be able to easily check the audit or access events, these event should be logged into separate files.

- What are audit logs: log statements produced when a user interacts with a resource, e.g. retrieving or updating configuration files, adding new users, login, ...
- What are access logs: when a agent is fetching a specific configuration

As a result of this ticket, the configuration server should generate a general log file, containing ALL log statements.
In addition, two log files should exist (e.g. `access.log` and `audit.log`) which contains only the respective log statements.

ArianeKraus:
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/522)


ArianeKraus:
image
image
image

PR #540, opened 2020-01-24
Closes #562 - Enable metrics to be written multiple times within a rule and scope

### LINKED ISSUE #562: Enable metrics to be written multiple times within a rule and scope.
Currently, a metric can be written only once within a rule on a certain scope. For certain circumstances it can be quite useful when a metric can be written multiple times on a specifc scope using different values.

JonasKunz:
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/540)


ivansenic:
@JonasKunz @mariusoe So I pushed small update to fix the review comments I had (since Jonas is on the vacation).

PR #546, opened 2020-01-29
Closes #545 - Config files listed in the config server UI can be moved using drag and drop

### LINKED ISSUE #545: File-Tree files of the config server UI should be moved using drag and drop
Currently, files/folders are moved using the corresponding buttons. Being able to use drag and drop would be more intuitive.

ArianeKraus:
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/546)


ArianeKraus:
UI bug :) the event isn't triggered

PR #561, opened 2020-02-14
Closes #556 - Obfuscation config available for trace attributes

### LINKED ISSUE #556: Posibility to obfuscate data pushed to traces
We need a configuration property for specifying patterns that should be used for checking if data pushed to the traces should be obfuscated or not. The pattern could be applied against the property name/key, value or both. 

The obfuscation should be performed against:
* span attributes
* span name?  

The pattern definition should specify the type of the pattern. Currently we have following matcher modes (should we use the same?):
- EQUALS_FULLY
- EQUALS_FULLY_IGNORE_CASE
- STARTS_WITH
- STARTS_WITH_IGNORE_CASE
- ENDS_WITH
- ENDS_WITH_IGNORE_CASE
- CONTAINS,
- CONTAINS_IGNORE_CASE
- MATCHES

Furthermore:
- meaningful defaults should be defined
- documentation must be updated

@JonasKunz @mariusoe Would be great if you could give answers to the above questions and propose the best place in the configuration model to have the settings available. As this could be applied to metrics later on, maybe it's better to have this as more generic option.

ivansenic:
Quick review intro..

What's missing:
* self monitoring -> if needed I would create a delegating `IObfuscatory` with self monitoring, quite easy
* documentation -> please suggest where should this be (what part, chapter and in what extent should I write this)

What's up for discussion:
- There are several `TODO`s left, I would like your opinion
- I am not sure if config part is as @JonasKunz has in mind, I am open for changes if necessary
- I did not implement the replacement for the obfuscated data per regex and just used `***`.. The reason is that my map shortcut would then need to be different (as I would need to not only remeber that the key should be obfuscated, but also what's the replacement).. Furthermore, as patterns can be defined against keys, and values are obfuscated, then the replacement would need to have two inputs (as in replaceAll()) method.. Up for discussion, as this way any data collected would have `***`, so why collect it at all.

What should be checked:
- I changed the Gradle build pick of the `JAVA_HOME` as it was not working when started from Intellij. Please check if system tests work from console + IDE for you

---
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/561)


ivansenic:
___
*[inspectit-ocelot-core/src/main/java/rocks/inspectit/ocelot/core/privacy/obfuscation/impl/PatternObfuscatory.java, line 18 at r1](https://reviewable.io/reviews/inspectit/inspectit-ocelot/561#-M0buz1B7R1Ymbhr7toK:-M0c6jzm6NbZT7p2g5sT:bnrnvjv) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/4c88f93b6965a721f76eb0fa37c246557fdd99e4/inspectit-ocelot-core/src/main/java/rocks/inspectit/ocelot/core/privacy/obfuscation/impl/PatternObfuscatory.java#L18)):*
<details><summary><i>Previously, mariusoe (Marius Oehler) wrote…</i></summary><blockquote>

Please clarify - I don't get your point.
</blockquote></details>

As we discussed no real benefit in splitting this..

PR #571, opened 2020-02-21
Closes #570 - Added tree table to configuration UI for showing primitive configuration properties

### LINKED ISSUE #570: Visualize the simple properties of the config file
The config file editing should allow possibility to show (in split-screen, or any other way) the plain properties that are defined in the file. This should be show in the a hierarchical tree-table. This is a first step into going for the editing of the plain properties on the UI.

ivansenic:
Enables viewing of the plain properties in the tree table view. The schema is pulled once the user is logged-in or when the state is rehydrated. The options to show or hide the UI view for the properties is saved in redux, so it becomes a general option.

Open questions and improvements:
* should we mark composite elements differently
* do we need a note that we only show the plain properties in the tree
* I did not implement filtering / searching / etc I think it would be too much for first commit

Looks like following:

image

Happy reviewing.. I suggest review first commit first, and then the second one, it's completly separate concerns..

---
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/571)


ivansenic:
*[a discussion](https://reviewable.io/reviews/inspectit/inspectit-ocelot/571#-M1zpnJP6H1qZ_Gxh4ic:-M1zpnJP6H1qZ_Gxh4id:b-fbl1sv) (no related file):*
Capital d in the Ocelot defaults

---

PR #622, opened 2020-03-13
Closes #580 - Edit of the properties (string, boolean, numbers) in the…

### LINKED ISSUE #580: Interface for editing the simple properties of the config file
The tree table implemented in the #570 should be extended for the editing mode of the non-composite properties. The edit should be reflected in the configuration file, if possible by only changing the single line where the value to change actually is. As alternative, the whole file could be flushed out.

ivansenic:
… tree table

**This is work in progress. Do not merge this until all the issues are resolved.**

So I am submitting the edit of the configuration view as it was originally implemented by me (without any suggestions from @JonasKunz  and others). I only adapted it to the integration changes that were done in the #570.

Current issues are:
* I am updating a single line in YAML. Of course, this is not the best solution, but it does keep the file content as before and does not remove comments. I am up in substituting this with the json to YAML dump. Let's just agree what should be done here wrt the information given to the user.
* Multiline string editing does not work. If we go with the dump this will be solvable easily (maybe I will add the text area instead). Otherwise, we could disable editing of the multiline strings.
* Durations and enums are nut supported atm. I could add durations support easily, but maybe in separate ticket to speed this out. Enums we don't have at the moment, so I am not sure if this should be added as well. Maybe also a separate ticket.
* A huge problem is the properties that values are corresponding to the other values. Especially in cases when the property type is number for example. So here for now when editing you can only replace it with the number, not the reference. Maybe this is a good solution, we could follow in other cases.
* Please respond to all the other `TODO` items that we not mentioned above

I would like us to define the minim set of things needed to be done & fixed in this ticket and then create other issues for additional features and improvements. Let's move fast.

Not sure if any documentation updates are needed, but I could not find any place where UI is mentioned or shown.

---
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/622)

### DIFF
--- a/components/inspectit-ocelot-configurationserver-ui/src/components/editor/EditorToolbar.js
+++ b/components/inspectit-ocelot-configurationserver-ui/src/components/editor/EditorToolbar.js
@@ -22,15 +22,15 @@ const EditorToolbar = ({
   <div className="this">
     <style jsx>
       {`
-      .this :global(.p-toolbar) {
+        .this :global(.p-toolbar) {
           border: 0;
           border-radius: 0;
           background-color: #eee;
           border-bottom: 1px solid #ddd;
-      }
-      .p-toolbar-group-right > :global(*) {
+        }
+        .p-toolbar-group-right > :global(*) {
           margin-left: 0.25rem;
-      }
+        }
         .this :global(.p-button-outlined) {
           color: #005b9f;
           background-color: rgba(0, 0, 0, 0);
--- a/components/inspectit-ocelot-configurationserver-ui/src/components/editor/EditorToolbar.js
+++ b/components/inspectit-ocelot-configurationserver-ui/src/components/editor/EditorToolbar.js
@@ -22,15 +22,15 @@ const EditorToolbar = ({
   <div className="this">
     <style jsx>
       {`
-        .this :global(.p-toolbar) {
+      .this :global(.p-toolbar) {
           border: 0;
           border-radius: 0;
           background-color: #eee;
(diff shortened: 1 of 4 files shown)


ivansenic:
@JonasKunz @mariusoe I updated the PR yesterday, this is what has been done:

* I moved to a new component structure and handling of the parsing and dumping in one central component - `VisualEditor`
* In the same editor we now centrally handle errors and show dialog to warn the users that visual editing will remove comments (I did not remove `Yes and remember`, that's one of the things to be done in next set of changes).
* I updated and now wrapped editing components with extra buttons `Null`, `NotNull`, `Set` & `Unset`
* I am showing all properties even the ones that are not in the config file. This is compatible with showing the props that are in the config file but not in schema.
* There is an option to show/hide non defined props in the file
* I fixed the bugs in the top menu and hid actions related to the Ace editor
* Regarding the editing:
   * numbers have normal input with key filter that matches numbers and properties substitution 
   * booleans have editable dropdown
   * durations use normal string input
   * enums are still not editable

The only problem is the editable dropdown for the boolean values. Editing does not work as focus is always lost after clicking on the input (blur event fired).. I tried to debug this, but failed to figure it out.. If you have time, would be great if somebody could invest like 30 mins to check if there is a quick solution for this.. Updating the primereact did not work.. I guess it could be connected to the editing events fired from the table itself.. Not sure.. But this is not working and we need to find a solution of agree on what to support now.

I also don't have support for multiline strings, will add this as well in the next round of changes..

PR #651, opened 2020-03-30
Closes #637 - Exposing count of reported resource timing elements

### LINKED ISSUE #637: Record EUM metrics of resource timings collected by Boomerang
Currently, the EUM server is not capabale of collecting resource timing metrics, which are collected by Boomerang's _restiming_-plugin.

The reason for this is that these metrics are not send as plain beacon properties but send in the encoded `restiming` beacon field. See the [plugin documentation](https://github.com/akamai/boomerang/blob/master/plugins/restiming.js) for more information on the encoding.

Example:
```
restiming=
{
  "http": {
    "://localhost:8080/": {
      "monitored-html.html": "6,7,7,6,4,,4,4,4*1qe,_",
      "boomerang/plugins/": {
        "r": {
          "t.js": "3v,e,5,4*118aj,_*20",
          "estiming.js": "3w,v,5,4*11c0f,_*20"
        },
        "navtiming.js": "3w,l,4,3*1c8a,_*20",
        "auto-xhr.js": "3y,t,b,2*11tq9,_*20"
      }
    },
    "s://www.google.de/images/branding/googlelogo/1x/googlelogo_color_272x92dp.png": "*02k,7k,1y,9s|1y,t"
  }
}
```

In this ticket, the EUM server should be extended to be able to collect the timing metrics of the monitored resources (see previous example) provided by the received Boomerang beacons.

ivansenic:
Few hits for the review:

* I think without reading a full docs from the Boomerang, it's impossible to review this, especially the parsing of the compressed string for resource timing
* As discussed with @JonasKunz I added a `BeaconRecorder` interfaces that acts on fully processed beacon, extended the manager for this purpose
* Currently I am only counting resources.. However, I would advise that we immediately report the load times instead. We will have the COUNT and SUM views, so we would have both the count on how many, plus the loading times.. Imo there is no reason to start only with counting.. Plus, let's agree on the metric name not sure if `resource_timing_total` is good one. Note that now I could not test the `timing` values, as these are not reported, thus even the tests would be more complete.
* I added the `crossOrigin` tag as well, think this is important. I report `cached` tag only if it's same origin, otherwise we can not know.
* There are few `TODO`s to be discussed
* Documentation comes on top when we agree on the final metrics

I am here for questions. Would be great if somebody can additionally test this on a meaningful HTMl page.

---
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/651)

### DIFF
--- a/components/inspectit-ocelot-eum-server/src/main/java/rocks/inspectit/oce/eum/server/beacon/recorder/BeaconRecorder.java
+++ b/components/inspectit-ocelot-eum-server/src/main/java/rocks/inspectit/oce/eum/server/beacon/recorder/BeaconRecorder.java
@@ -1,0 +1,18 @@
+package rocks.inspectit.oce.eum.server.beacon.recorder;
+
+import rocks.inspectit.oce.eum.server.beacon.Beacon;
+
+/**
+ * Interface for all components acting as {@link BeaconRecorder}.
+ * BeaconRecorder are intended to record custom complicated metrics from a fully-processed Beacon.
+ */
+public interface BeaconRecorder {
+
+    /**
+     * Records arbitrary metrics from given {@link Beacon}
+     *
+     * @param beacon Fully-processed beacon
+     */
+    void record(Beacon beacon);
+
+}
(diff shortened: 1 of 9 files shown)


ivansenic:
@mariusoe I pushed the updates..

However, there are two `TODO`s left that Jonas did not respond to.. In addition, I have currently defined the metric as the `resource_time`, but as type long and I am adding `1.0` as time.. This all seems not correct imo.. Let's shortly talk after you check the `TODO`s maybe.. I also added documentation, so you could have a look on that as well..

PR #668, opened 2020-04-09
Closes #391 - Handle invalid metric tags

### LINKED ISSUE #391: Handle invalid metric tags in agent
Currently, if the agent tries to tag a metric with a tag value with is not considered as valid (see https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/tags/TagValue.java) it will throw an exception and the hook which is recording the metric will be deactivated.

This should be handled, e.g. the tag value could be set to `<invalid>`. By doing this, bad configurations are also be visible using the metrics.

arolfes:
util method in CommonTagsManager to create TagValue
for all invalid TagValues is <invalid> generated

---
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/668)


arolfes:
there are also 3 usages of `TagValue.create` in component `inspectit-ocelot-eum-server`
should code be changed also?

arolfes:
I faced this problem on my machine also
but run the tests a second time the error disappeared
i have no idea why

PR #673, opened 2020-04-13
Closes #602 - Option to use pre authentication on the configuration se…

### LINKED ISSUE #602: Implement proxy authentication for the configuration server
For easier usage of the configuration server's web-UI, the configuration server should be able to support a proxy authentication.

If the config server is used behind a reverse proxy, which is able to authenticate the current user, it should be possible to pass this information to the configuration server.
The config server should be able to login a user which is specified in a specific header field, set by the reverse proxy.

This function should be similar to the "Auth Proxy" function of Grafana. See: https://grafana.com/docs/grafana/latest/auth/auth-proxy/

ivansenic:
So many things to discuss:

* First, there is security config switch now, if proxy auth is enabled, most of the existing stuff is dismissed:
   * There is only pre authentication, so no token, basic, ldap.. This is fine imo and has to be like this..
   * However if ldap is active we would use ldap to still retrieve the details about the user, thus correctly resolve the roles..
   * Some config is shared and this is extracted into a super class
* Works as simple as sending the username in the header, from there on we consult the active user service to resolve if user exists and what are his roles & grants..
* It works only if the username is passed to the existing user of course..

From the UI point (did not touch a thing until we agree):
* Login form is not needed any more when the auth proxy is on
  * So what we can do is try to fetch the token when the login form is shown in order to see if the token can be fetched without the auth (so using the proxy auth)
  * If this works everything else can be the same.. Token that is received would be only used on the UI, it would have no effect on the requests made as you can not auth with the token in this scenario
  * I guess this is how we can have minimum changes on the UI to be able to support both auth scenarios

However, it does not make sense to have a login form at all on the UI. We should have Oauth support and then we would not care what mode the backend is on. If it's default stuff, then the backend will present the login form and dispatch the oauth code on successful login that can be used to get the token. If it's auth proxy you would directly get the code.. The UI would not care.. However, this requires a lot of adaptation on the backend and the frontend.. I am not sure why this was not implemented like this in the first place?

In addition, the token should be used to access the api and the pre-auth or form auth only to retrieve the token. This is how correctly this should be split imo.

---
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/673)

### DIFF
--- a/components/inspectit-ocelot-configurationserver/src/main/java/rocks/inspectit/ocelot/config/model/AuthProxySettings.java
+++ b/components/inspectit-ocelot-configurationserver/src/main/java/rocks/inspectit/ocelot/config/model/AuthProxySettings.java
@@ -1,0 +1,27 @@
+package rocks.inspectit.ocelot.config.model;
+
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+/**
+ * Settings for the configuration of the to let a HTTP reverse proxy handling authentication.
+ */
+@Data
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor
+public class AuthProxySettings {
+
+    /**
+     * If auth proxy is enabled.
+     */
+    private boolean enabled;
+
+    /**
+     * The request header where the username is stored.
+     */
+    private String principalRequestHeader;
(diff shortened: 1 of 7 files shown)


quandor:
If, after almost three years, we ever pick this up, we start from scratch.

PR #679, opened 2020-04-15
Closes #669 - Added percentile views

### LINKED ISSUE #669: Add Percentile Views to the Agent
* The agent should have a dedicated view Type for computing percentils and min / max
* For the resulting views, sliding windows of data will be stored
* Resulting metrics are exposed as gauges via OpenCensus

JonasKunz:
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/679)


ivansenic:
___
*[inspectit-ocelot-core/src/main/java/rocks/inspectit/ocelot/core/metrics/percentiles/AsyncPercentileRecorder.java, line 45 at r1](https://reviewable.io/reviews/inspectit/inspectit-ocelot/679#-M51GQ_j6bMmg3r1sAwC:-M51Zx9-29UONEeKc-Xs:bifj281) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/3862f349e04dfb54acd937690a774450dc1bb56b/inspectit-ocelot-core/src/main/java/rocks/inspectit/ocelot/core/metrics/percentiles/AsyncPercentileRecorder.java#L45)):*
<details><summary><i>Previously, ivansenic (Ivan Senic) wrote…</i></summary><blockquote>

So is this queue here only to support sync recording, meaning that the single thread is reading from the queue? Or just to have the writing as fast as possible?

I was looking into following impl, with higher performance and a nice drain method with a consumer:

https://github.com/real-logic/agrona/blob/master/agrona/src/main/java/org/agrona/concurrent/ManyToOneConcurrentArrayQueue.java

Not sure if something like this would be helpful here.. Would be nice if you could specify draining strategy somehow here :) I mean the more conservative you are, you will tend to miss inserting values in the current time window, the more aggressive you are you will be using CPU more..

Did you at all think about this missed values? so you insert it in the buffer too late for the current sliding window, but then you ignore it in the next one as timestamp is too old.. You could easily miss a value and thus have inappropriate results. I think this is easily testable?
</blockquote></details>

OK I think I made a mistake you always report percentiles for the last 15s from asking? This way you can not "lose" a value as I described above right..

ivansenic:
___
*[inspectit-ocelot-core/src/main/java/rocks/inspectit/ocelot/core/metrics/percentiles/PercentileView.java, line 36 at r3](https://reviewable.io/reviews/inspectit/inspectit-ocelot/679#-M5QrU94Asqm-vb1FnKy:-M5QrU94Asqm-vb1FnKz:b-7eovn) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/eebd56760ea4f0836afc2fe35bfb55459c16175c/inspectit-ocelot-core/src/main/java/rocks/inspectit/ocelot/core/metrics/percentiles/PercentileView.java#L36)):*
> ```Java
>      * The tag value to use for {@link #PERCENTILE_TAG_KEY} for the "minimum" series.
>      */
>     private static final String MIN_METRIC_SUFFIX = "_min";
> ```

should we have suffixes as `/min` to follow the OC standards.. Of course this would be translated to the `_min` in prometheus exporter? How is with sum/count?

PR #687, opened 2020-04-17
Closes #639 - Logback.xml in the configuration server JAR prevents cu…

### LINKED ISSUE #639: Logback.xml in the configuration server JAR prevents custom log configurations
Currently, the configuration server's JAR contains a `logback.xml` file. This is quite handy, because the server is shipped including a nice log format.

Unfortunately, this prevents users from using their custom `logback.xml` file. Defining custom `logback.xml` file using the `loggin.config` or `logback.configurationFile` property is not working but the bundle one is used.

This has to be fixed in order to be able to use external log configurations.

arolfes:
…stom log configurations

first try to read logback.xml from application property logback.configurationFile
second try to read logback.xml from application property logging.config
third try to read logging config from octelot config
fourth read logback.xml from classpath

---
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/687)


arolfes:
The issue is only about configuration-server.
I close this PR

PR #704, opened 2020-04-27
Closes #518 - Refactoring the file access in the configuration server

### LINKED ISSUE #518: Refactor the FileManager with a concept of "Views"
The idea is to simplify the usage of the `FileManager` by introducing the concept of `DirectoryView`s.
`DirectoryView` could be a ninterface with the following methods:
```
interface DirectoryView {
  readFile(String path)
  doesExist(String path)
  listFiles(String path, boolean recursive)
  isDirectory(String path)
}
```

The FileManager would now expose two methods returning different Implementations of such Views:
`workingDirectory()` and `versioned(String version)`. The working-directory implementation would allow the editing of files in addition whereas the versioned implementation is backed by git and exposes additional information such as the commit date or the author.

In addition the FIleManager has functionality for querying available versions, commiting, performing a rollback, etc..

Why would this be beneficial?
Firstly many places of file access actually don't care if it's the working-directory or a versioned one, like the Agent-Configuration-loader or potential future implementations for doing server side validation. This makes such code less error prone. It avoids for example accidently doing a `doesExist()` on the working-directory while actually reading files from a specific version.

In addition it avoids race conditions when reading files from the current latest version, because a view would always refer to a fixed version, even if in the mean-time a new commit to the master branch happens.

mariusoe:
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/704)


ivansenic:
___
*[components/inspectit-ocelot-configurationserver/src/main/java/rocks/inspectit/ocelot/file/accessor/AbstractFileAccessor.java, line 105 at r1](https://reviewable.io/reviews/inspectit/inspectit-ocelot/704#-M63yKUM5grwmtUVfdzf:-M6TkNriAzJzbMcdUqyk:b32w0x) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/d2dd378cd4f603f48e3dd00c9cabfa21388b776f/components/inspectit-ocelot-configurationserver/src/main/java/rocks/inspectit/ocelot/file/accessor/AbstractFileAccessor.java#L105)):*
<details><summary><i>Previously, mariusoe (Marius Oehler) wrote…</i></summary><blockquote>

I would keep it with the concret definition. Imo its easier to understand and you ensure, that the user specifies two parameters and not only one or maybe more than 3.
</blockquote></details>

OK

PR #718, opened 2020-05-05
Closes #717 - Deleting config files using keyboard

### LINKED ISSUE #717: Deleting config files using keyboard
As normal editors usually allow for deleting files by pressing the delete key, it would be nice having this as well.

ArianeKraus:
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/718)


mariusoe:
Can you're tests are failing because the files are not formatted correctly.

Run `yarn lint:fix` and `format:write` to show the problems, and solve the ones which can be solved automatically.

PR #760, opened 2020-05-27
Add 2end Copy from boomerang.js with the Name inspectit.js

### LINKED ISSUE #748: EUM / boomerang.js blocked by ublock Origin EasyPrivacy list
If one has AdBlocker "ublock" activated the "easyprivacy list" is blocking "/boomerang.js"
There should be a way to set a alias name for boomerang.js

mc-chaos:
Close #748

---
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/760)

### DIFF
--- a/components/inspectit-ocelot-eum-server/build.gradle
+++ b/components/inspectit-ocelot-eum-server/build.gradle
@@ -40,6 +40,12 @@ bootJar {
 		include "boomerang.js"
 		into "static/boomerang"
 	}
+	from ("$buildDir/boomerang-${boomerangVersion}"){
+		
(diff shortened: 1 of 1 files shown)


mariusoe:
Thanks for the contribution @mc-chaos 

But we think that providing multiple boomerang files with different names is not the best solution for this problem. Because we have the same file(s) multiple time and we are again bound to the specified name.
We would rather do it the way you already suggested in the ticket - providing a configuration option to set an alias for the file.

mc-chaos:
We will go an other way

PR #771, opened 2020-06-03
Creating hooks for event generation

ArianeKraus:
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/771)


ArianeKraus:
___
*[inspectit-ocelot-core/src/main/java/rocks/inspectit/ocelot/core/instrumentation/hook/MethodHookGenerator.java, line 221 at r1](https://reviewable.io/reviews/inspectit/inspectit-ocelot/771#-MBY3BZj1PoMT9QbMNrC:-MBYYUaX8EFZVrBFrv5Q:b-v76f2z) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/7ae2d94aff8165f09a30e32092a1614be4d8371e/inspectit-ocelot-core/src/main/java/rocks/inspectit/ocelot/core/instrumentation/hook/MethodHookGenerator.java#L221)):*

Why is the event-name a data-key?
</blockquote></details>

I found it useful being able to provide the resulting event name, with which the event will be exported to be provided as a variable as well... Since it is possible to define one event for multiple scopes it should be possible to differentiate them right away.

More of an example: I've got multiple scopes where each should create the same event with attributes. The only difference apart from the values is the event name, which is (in my case) based on the called method name. So in this case the event name can be provided as a data key and the difference would have been processign the events afterwards or defining a new rule for each scope

ArianeKraus:
___
*[inspectit-ocelot-core/src/main/java/rocks/inspectit/ocelot/core/instrumentation/hook/MethodHookGenerator.java, line 239 at r1](https://reviewable.io/reviews/inspectit/inspectit-ocelot/771#-MBY3Hq91cHjNWugruuC:-MBYaTQCEtkUMEFp5CNE:b-b9y98o) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/7ae2d94aff8165f09a30e32092a1614be4d8371e/inspectit-ocelot-core/src/main/java/rocks/inspectit/ocelot/core/instrumentation/hook/MethodHookGenerator.java#L239)):*

What about lists?
</blockquote></details>

Later :)

PR #779, opened 2020-06-10
Develop

bunckelvhv:
Support for JBoss Logmanager. Native Logger of JBoss EAP and WildFly. https://github.com/jboss-logging/jboss-logmanager/blob/master/core/src/main/java/org/jboss/logmanager/MDC.java

---
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/779)


mariusoe:
This change has been included in PR #800 and #799

PR #860, opened 2020-07-20
Finishing Functionality for Basic Version of Event Exporter

ArianeKraus:
Added EventExporterService + RegistryService for exporter handlers
Resolving events and exporting them

- only maps (not lists) are resolved.


quandor:
If, after almost three years, we ever pick this up, we start from scratch.

PR #899, opened 2020-08-05
Closes #482 - Configuration Reload can happen while the Agent is initializing

### LINKED ISSUE #482: Configuration Reload can happen while the Agent is initializing
Ocelot offers ways of dynamically reloading configurations (HTTP, file-based). This however should only happen after the spring context has been fully initialized, which is not the case at the moment.

This potentially could lead to race conditions.

mariusoe:
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/899)


JonasKunz:
Closed in favor of #901

PR #1000, opened 2021-02-10
Closes #973 - Context correlation will not swap instances to prevent application errors

### LINKED ISSUE #973: Possible class cast exception in custom Executor implementations
In the current implementation, the agent wraps a `Runnable` instances for trace correlation when it is passed to a `Executor` or `ScheduledExecutorService`. Currently, an anonymous class will be created for this. In case the `Executor` implementation knows the exact class of the passed `Runnable` and tries to cast it to that class, an exception will occur because the class has been changed due to the agent's wrapping. E.g. this happens in [Glassfish's ManagedFutureTask](https://github.com/eclipse-ee4j/concurrency-ri/blob/ac7d098c0fab6849c9ed28b2b20c4a2580d2135a/src/main/java/org/glassfish/enterprise/concurrent/internal/ManagedFutureTask.java#L75)

This problem has to be fixed, because it can affect the application itself. In order to solve this, the anonymous class has been replaced by a derived class of the actual `Runnable` class, so that the wrapped class is compatible to the original one.

mariusoe:
closes #973

---
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1000)


Patrick-Eichhorn (approved):
> Reviewed 26 of 26 files at r1.
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1000)* status: :shipit: complete! all files reviewed, all discussions resolved

mariusoe:
PR has been reverted. Please see PR #1010

PR #1030, opened 2021-03-05
Introduce UI Editor for EUM-Server

MariusBrill:
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1030)


SaschaBrechmannVHV:
It looks like, this UI is not integrated in inspect-IT configsrv?

MariusBrill:
Hi @SaschaBrechmannVHV ,

This is just an experimental function that we needed and will not be included in the master branch (for now), hence the extra branch. We see it as separate from the configuration server since the EUM-Server could also be used without the configuration server. However, this feature could also be integrated in a future version of the configuration server. This would also mean we would have to adress the issues you mentioned (Auth, Groups, Roles etc.) before integrating it into the config server.

Greetings
Marius

PR #1094, opened 2021-05-28
Closes #1073 - Implement the functionality to remove the graphical configuration editor from configurations

### LINKED ISSUE #1073: Implement the functionality to remove the graphical configuration editor from configurations
In this ticket, the graphical configuration view of the configuration UI is to be extended so that a button exists with which the configuration can be converted into a pure YAML configuration. This means that users can "remove" the graphical editor - which means that the header comment is deleted. This allows users to create a base configuration using the graphical editor and extend the configuration manually using the raw YAML view.

When clicking this button, a warning should be show, stating that this operation cannot be undone.

image

Patrick-Eichhorn:
closes #1073

---
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1094)


mariusoe:
this pr is included in #1105

PR #1111, opened 2021-06-16
Closes #1107 - ScopeWizard is not correctly handling the method argument matcher

### LINKED ISSUE #1107: ScopeWizard is not correctly handling the method argument matcher
In the current version of the scope-wizard, the wizard is not handling the argument matcher correclty.
E.g. it is not removing the argument matcher in case the argument matching is disabled. Furthermore, when selecting a method without any argument via the class browser, the scope-wizard uses a single and empty argument, which is wrong in that case.

mariusoe:
closes #1107

---
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1111)

### DIFF
--- a/components/inspectit-ocelot-configurationserver-ui/src/components/editor/method-configuration-editor/MethodConfigurationEditor.js
+++ b/components/inspectit-ocelot-configurationserver-ui/src/components/editor/method-configuration-editor/MethodConfigurationEditor.js
@@ -17,7 +17,7 @@ import { selectedFileContentsChanged } from '../../../redux/ducks/configuration/
 import { useDispatch } from 'react-redux';
 
 /** data */
-import { CONFIGURATION_TYPES, TOOLTIP_OPTIONS } from '../../../data/constants';
+import { TOOLTIP_OPTIONS } from '../../../data/constants';
 
 const SCOPE_STATES_RULES = {
   TRACING: 'r_method_configuration_trace',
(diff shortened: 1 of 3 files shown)


Patrick-Eichhorn (changes requested):
> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1111)* status: all files reviewed, 1 unresolved discussion (waiting on @mariusoe)

*[a discussion](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1111#-McJ684k1UPzOCUrbKal:-McJ684k1UPzOCUrbKam:b-7i6rh7) (no related file):*
This implementation solves the problem  `“E.g. it is not removing the argument matcher in case the argument matching is disabled”`.

Do you want to solve the problem `"Furthermore, when selecting a method without any argument via the class browser, the scope-wizard uses a single and empty argument, which is wrong in that case"` in this PR as well?

---

mariusoe (commented):
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1111)* status: all files reviewed, 1 unresolved discussion (waiting on @Patrick-Eichhorn)

*[a discussion](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1111#-McJ684k1UPzOCUrbKal:-McONK5wCOGIFzlWrn1Q:bndi1es) (no related file):*
<details><summary><i>Previously, Patrick-Eichhorn wrote…</i></summary><blockquote>

This implementation solves the problem  `“E.g. it is not removing the argument matcher in case the argument matching is disabled”`.

Do you want to solve the problem `"Furthermore, when selecting a method without any argument via the class browser, the scope-wizard uses a single and empty argument, which is wrong in that case"` in this PR as well?
</blockquote></details>

Yes, the second problem is also solved using the following code `const parameters = match[3].split(', ').filter((value) => !!value);` in the scope-wizard.

The problem was that the `split` function returns an array containing a single, empty string in case the input is empty. Now, all empty elements are removed before using them.

---

Patrick-Eichhorn (approved):
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1111)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mariusoe)

Issue #1163, opened 2021-07-28
Replace Stream.peek with forEach


boris-unckel:
Stream.peak is for debugging purposes: [java.util.stream.Stream#peak](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#peek-java.util.function.Consumer-)
It shall be replaced with foreach

PR #1164, opened 2021-07-28
Closes (#1163) Replace Stream.peek with map

boris-unckel:
Fix https://github.com/inspectIT/inspectit-ocelot/issues/1163

---
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1164)

### DIFF
--- a/components/inspectit-ocelot-configurationserver/src/main/java/rocks/inspectit/ocelot/file/versioning/VersioningManager.java
+++ b/components/inspectit-ocelot-configurationserver/src/main/java/rocks/inspectit/ocelot/file/versioning/VersioningManager.java
@@ -444,11 +444,7 @@ public class VersioningManager {
         List<SimpleDiffEntry> simpleDiffEntries = diffEntries.stream()
                 .map(SimpleDiffEntry::of)
                 .filter(entry -> entry.getFile().startsWith(AbstractFileAccessor.CONFIGURATION_FILES_SUBFOLDER))
-                .peek(entry -> {
-                    String shortenFile = entry.getFile()
-                            .substring(AbstractFileAccessor.CONFIGURATION_FILES_SUBFOLDER.length());
-                    entry.setFile(shortenFile);
-                })
+                .map(SimpleDiffEntry::shortenName)
                 .collect(Collectors.toList());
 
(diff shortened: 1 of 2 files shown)


components/inspectit-ocelot-configurationserver/src/main/java/rocks/inspectit/ocelot/file/versioning/VersioningManager.java
```diff
@@ -444,7 +444,7 @@ WorkspaceDiff getWorkspaceDiff(boolean includeFileContent, ObjectId oldCommit, O
         List<SimpleDiffEntry> simpleDiffEntries = diffEntries.stream()
                 .map(SimpleDiffEntry::of)
                 .filter(entry -> entry.getFile().startsWith(AbstractFileAccessor.CONFIGURATION_FILES_SUBFOLDER))
-                .peek(entry -> {
+                .forEach(entry -> {
```
mariusoe, line 447: This would not work because `forEach` terminates the stream, thus, the elements cannot be collected in a list.
If we want to replace `peek` we might have to use `map` and return the `entry` object after it has been modified. Maybe this would be the better way and by doing this, we'll align to the Javadocs and don't mutate the elements using `peek`.

mariusoe (changes requested):

components/inspectit-ocelot-configurationserver/src/main/java/rocks/inspectit/ocelot/file/versioning/VersioningManager.java
boris-unckel: @mariusoe I have changed to `map` . Please check whether SimpleDiffEntry is the right place for the util method.

mariusoe (approved):
:lgtm:

> Reviewed 1 of 2 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1164#-:-MkCTYmx8ZcSzjXv427J:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @boris-unckel)

PR #1196, opened 2021-08-19
Closes #1195 - Change the agent's transformation approach to decorateinstead of redefine

### LINKED ISSUE #1195: Change the agent's transformation approach to `decorate`instead of `redefine`
Currently, the agent uses ByteBuddy and its `redefine` approach for generating an instrumented bytecode for a specific class. In this case, ByteBuddy loads and requests a lot of data which is needed for its redefinition. Due to the fact, that the agent only decorates method and does not add or change any method or fields, a lot of work done by the `redefine`method is not needed. Furthermore, loading all this additional data can lead to some problems when instrumenting certain classes. See: https://github.com/raphw/byte-buddy/issues/1095 and https://github.com/raphw/byte-buddy/issues/1040

In this ticket, the approach for generating bytecode should be switched to ByteBuddys `decorate` and its method graph resolution to `ForDeclaredMethods` which should be sufficient for us. This should solve problems where the bytecode is not completly valid (see mentioned ticket).
As a side effect, this approach should be slightly faster than the current one.

mariusoe:
closes #1195

---
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1196)

### DIFF
--- a/inspectit-ocelot-core/src/main/java/rocks/inspectit/ocelot/core/instrumentation/AsyncClassTransformer.java
+++ b/inspectit-ocelot-core/src/main/java/rocks/inspectit/ocelot/core/instrumentation/AsyncClassTransformer.java
@@ -11,6 +11,7 @@ import net.bytebuddy.description.method.MethodDescription;
 import net.bytebuddy.description.type.TypeDescription;
 import net.bytebuddy.dynamic.ClassFileLocator;
 import net.bytebuddy.dynamic.DynamicType;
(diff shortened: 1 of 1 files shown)


Patrick-Eichhorn (approved):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1196)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mariusoe)

PR #1206, opened 2021-09-06
Closes #1205 - Reading configuration from mapped Kubernetes Config Maps

### LINKED ISSUE #1205: [Configuration Server] java.io.FileNotFoundException in FileInfoVisitor when trying to read from a directory that is referenced by a sym link
Hello inspectIT Team,
I stumbled over this bug while running the inspectIT Ocelot Configuration Service in Kubernetes and passed some instrumentation yaml as Config Map. The challenge is that Kubernetes uses symlinks within the filesystem as seen here:

root@inspectit-ocelot-configuration-server-6d58469594-5kk5z:/configuration-server/files# ls -lah
...
drwxr-xr-x 2 root root 4.0K Sep  1 12:15 ..2021_09_01_12_15_55.274058932
lrwxrwxrwx 1 root root   31 Sep  1 12:15 ..data -> ..2021_09_01_12_15_55.274058932
lrwxrwxrwx 1 root root   16 Sep  1 12:15 some_instrumentation.yml -> ..data/some_instrumentation.yml

The problem now is that the FileInfoVisitor tries to read the sym link '..data' that references a folder and not a file. In addition, the mysql.yml would not be read as the sym link is not followed.

I tried providing a fix in commit https://github.com/matthias-huber/inspectit-ocelot/commit/14b9783053afd4d15decb75d5ff6fc7bdf764d02
The implementation skips hidden folders (otherwise the some_instrumentation would be loaded twice) and follows sym-links.
 The tests are not covering all cases right now and don't validate the successful reads.

Just wanted to inform you on the issues with sym links (any maybe hidden folders) when using Config Maps in k8s. Most likely the use of Config Maps for instrumentation is not used for production but rather for demo or fast initial setups as I think it doesn't work together with the idea of keeping the instrumentation under version control.

matthias-huber:
Kubernetes uses symlinks within the filesystem to allow updates of Config Maps during runtime.
This fix skips hidden folders with configuration files and symbolic links are followed and read from
in case that a file is referenced by the symbolic link.

---
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1206)

### DIFF
--- a/components/inspectit-ocelot-configurationserver/src/main/java/rocks/inspectit/ocelot/file/FileInfoVisitor.java
+++ b/components/inspectit-ocelot-configurationserver/src/main/java/rocks/inspectit/ocelot/file/FileInfoVisitor.java
@@ -14,6 +14,7 @@ import java.io.IOException;
 import java.lang.reflect.Type;
 import java.nio.file.FileVisitResult;
 import java.nio.file.FileVisitor;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.util.*;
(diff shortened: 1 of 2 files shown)


mariusoe (approved):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1206)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matthias-huber)

PR #1346, opened 2022-03-07
fix(EUM-server): Add conversion-service to EUM server

aaronweissler:
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1346)

### DIFF
--- a/components/inspectit-ocelot-eum-server/src/main/java/rocks/inspectit/oce/eum/server/configuration/BeanConfiguration.java
+++ b/components/inspectit-ocelot-eum-server/src/main/java/rocks/inspectit/oce/eum/server/configuration/BeanConfiguration.java
@@ -5,6 +5,8 @@ import io.opencensus.stats.StatsRecorder;
 import io.opencensus.stats.ViewManager;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.core.convert.ConversionService;
+import rocks.inspectit.ocelot.config.conversion.InspectitConfigConversionService;
 
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
@@ -37,5 +39,10 @@ public class BeanConfiguration {
     @Bean
     public ScheduledExecutorService scheduledExecutor() {
         return Executors.newScheduledThreadPool(4);
+    }
+
+    @Bean("conversionService")
+    public ConversionService getConversionService() {
+        return InspectitConfigConversionService.getInstance();
     }
 }
--- a/components/inspectit-ocelot-eum-server/src/main/java/rocks/inspectit/oce/eum/server/exporters/beacon/BeaconHttpExporter.java
+++ b/components/inspectit-ocelot-eum-server/src/main/java/rocks/inspectit/oce/eum/server/exporters/beacon/BeaconHttpExporter.java
@@ -3,6 +3,7 @@ package rocks.inspectit.oce.eum.server.exporters.beacon;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.stereotype.Component;
 import rocks.inspectit.oce.eum.server.beacon.Beacon;
 import rocks.inspectit.oce.eum.server.configuration.model.BeaconHttpExporterSettings;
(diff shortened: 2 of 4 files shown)


heiko-holz (changes requested):
> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1346)* status: all files reviewed, 1 unresolved discussion (waiting on @aaronweissler)

___
*[components/inspectit-ocelot-eum-server/src/main/java/rocks/inspectit/oce/eum/server/exporters/beacon/BeaconHttpExporter.java, line 22 at r1](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1346#-Mx_6J-uDD8BYZHigFgg:-Mx_6J-uDD8BYZHigFgh:bmpfojk) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/ad3b06475bb189a37df171d4d512505ba7c96b39/components/inspectit-ocelot-eum-server/src/main/java/rocks/inspectit/oce/eum/server/exporters/beacon/BeaconHttpExporter.java#L22)):*
> ```Java
> @Slf4j
> @ConditionalOnProperty({"inspectit-eum-server.exporters.beacons.http.enabled"})
> @ConditionalOnExpression("NOT new String('${inspectit-eum-server.exporters.beacons.http.enabled}').toUpperCase().equals(T(rocks.inspectit.ocelot.config.model.exporters.ExporterEnabledState).DISABLED.toString())")
> ```

why do we need to convert it to `toUpperCase()`? 
Is the value parsed from deprecated configurations `disabled`, `DISABLED` (according to `BooleanToExporterEnabledStateConverter.java`), or `false`?

We could potentially also try to parse the `ExporterEnabledState` from the String and checker whether it is `.disabled()`, i.e. `ExporterEnabledState.valueOf(new String('${inspectit-eum-server.exporters.beacons.http.enabled}')).isDisabled()`

aaronweissler (commented):
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1346)* status: all files reviewed, 1 unresolved discussion (waiting on @heiko-holz)

___
*[components/inspectit-ocelot-eum-server/src/main/java/rocks/inspectit/oce/eum/server/exporters/beacon/BeaconHttpExporter.java, line 22 at r1](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1346#-Mx_6J-uDD8BYZHigFgg:-MxcyzVP0zlLIFTpgHWk:bm9nfj8) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/ad3b06475bb189a37df171d4d512505ba7c96b39/components/inspectit-ocelot-eum-server/src/main/java/rocks/inspectit/oce/eum/server/exporters/beacon/BeaconHttpExporter.java#L22)):*

why do we need to convert it to `toUpperCase()`? 
Is the value parsed from deprecated configurations `disabled`, `DISABLED` (according to `BooleanToExporterEnabledStateConverter.java`), or `false`?

We could potentially also try to parse the `ExporterEnabledState` from the String and checker whether it is `.disabled()`, i.e. `ExporterEnabledState.valueOf(new String('${inspectit-eum-server.exporters.beacons.http.enabled}')).isDisabled()`


</blockquote></details>

In both cases the problem is that this evaluation seems to happen really early on the raw input data before Spring uses the converters or even parses it.

The toUpperCase() is then needed, because someone could put in "disabled" as the `enabled` value which Spring would accept just fine and later parse as ExporterEnabledState.DISABLED, but in the evaluation it is still seen as "disabled" which would lead to the Component being built despite actually being supposed to be disabled. This part is a correction for the new correct way of setting `enabled`.

The deprecated way would be setting it to false, which is filtered by the `ConditionalOnProperty` annotation already.

This early evaluation also makes the recommendation impossible unfortunately, because if the old deprecated way is used this would throw an Exception, since it would try to parse "false" or "true" with valueOf, because they will only get converted correctly afterwards.

heiko-holz (approved):
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1346)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aaronweissler)

___
*[components/inspectit-ocelot-eum-server/src/main/java/rocks/inspectit/oce/eum/server/exporters/beacon/BeaconHttpExporter.java, line 22 at r1](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1346#-Mx_6J-uDD8BYZHigFgg:-MxhXuLu4XCVEgBKpQZi:b-91p33y) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/ad3b06475bb189a37df171d4d512505ba7c96b39/components/inspectit-ocelot-eum-server/src/main/java/rocks/inspectit/oce/eum/server/exporters/beacon/BeaconHttpExporter.java#L22)):*

In both cases the problem is that this evaluation seems to happen really early on the raw input data before Spring uses the converters or even parses it.

The toUpperCase() is then needed, because someone could put in "disabled" as the `enabled` value which Spring would accept just fine and later parse as ExporterEnabledState.DISABLED, but in the evaluation it is still seen as "disabled" which would lead to the Component being built despite actually being supposed to be disabled. This part is a correction for the new correct way of setting `enabled`.

The deprecated way would be setting it to false, which is filtered by the `ConditionalOnProperty` annotation already.

This early evaluation also makes the recommendation impossible unfortunately, because if the old deprecated way is used this would throw an Exception, since it would try to parse "false" or "true" with valueOf, because they will only get converted correctly afterwards.
</blockquote></details>

thanks for the exhaustive explanation!

PR #1348, opened 2022-03-11
Update the agent image to be compatible with the OTel K8s operator

aaronweissler:
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1348)


aaronweissler (commented):
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1348)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[inspectit-ocelot-agent/docker/Dockerfile, line 2 at r1](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1348#-Mxt6DmvDDFEwZuUbUvI:-Mxt6DmvDDFEwZuUbUvJ:bq0mdix) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/76f4271a18369fa4e282a0e466c476251ef574c9/inspectit-ocelot-agent/docker/Dockerfile#L2)):*
> ```Dockerfile
> FROM alpine:3
> ADD https://github.com/inspectIT/inspectit-ocelot/releases/download/1.15.1/inspectit-ocelot-agent-1.15.1.jar /javaagent.jar
> ```

Should/can this be set dynamically somehow instead of hardcoded version?

SaschaBrechmannVHV:
> ```dockerfile
> FROM alpine:3
> ```

Why do not start with an empty image like
`FROM scratch`
Do this container need any thing from alpine ? 
Regrads, Sascha

mariusoe:
Hi,
yes, as far as I can see, it should be sufficient using the `scratch` base image as @SaschaBrechmannVHV suggested. Currently, the container is doing nothing, so that should be fine - I would say..

> Should/can this be set dynamically somehow instead of hardcoded version?
Yes, absolutly. The agent used in the image should not be hard-coded. I would say we should use a local one there, so that you can use a specific one (e.g. a newly built one in a CI pipeline which is not online yet) when building the image.

aaronweissler (commented):
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1348)* status: 0 of 4 files reviewed, 1 unresolved discussion

___
*[inspectit-ocelot-agent/docker/Dockerfile, line 2 at r1](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1348#-Mxt6DmvDDFEwZuUbUvI:-My77DTp7PTgvRGLwkiJ:b-9u6vqc) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/76f4271a18369fa4e282a0e466c476251ef574c9/inspectit-ocelot-agent/docker/Dockerfile#L2)):*

Should/can this be set dynamically somehow instead of hardcoded version?
</blockquote></details>

Updated to use local jar.
___
*[inspectit-ocelot-agent/docker/Dockerfile, line 1 at r2](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1348#-My76Yf59mTu9WMylqbK:-My76Yf59mTu9WMylqbL:bopdrno) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/0f50f61cc5312c2820b78ad554ae4cad42fedb73/inspectit-ocelot-agent/docker/Dockerfile#L1)):*
> ```Dockerfile
> FROM busybox
> ```

@saschabrechmannvhv @mariusoe 
Using `scratch` is not possible because it does not provide the commands we need like `chmod`, `ln` or `cp`.
We can however use busybox which is also a bit smaller than alpine and should provide enough functionality for our use case (it is also the base image opentelemetry use for this purpose).
___
*[inspectit-ocelot-agent/docker/Dockerfile, line 5 at r3](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1348#-My77P6X0FjUWWWo8u6a:-My77P6X0FjUWWWo8u6b:b9lmak9) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/70f94b2302f0306ef3f63d14c87acb91faa7c621/inspectit-ocelot-agent/docker/Dockerfile#L5)):*
> ```Dockerfile
> COPY inspectit-ocelot-agent.jar /
> RUN ln /inspectit-ocelot-agent.jar /javaagent.jar &&\
>     chmod -R go+r /javaagent.jar
> ```

Combined the two RUN statements into one to reduce number of layers.
___
*[inspectit-ocelot-documentation/docs/getting-started/installation.md, line 96 at r3](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1348#-My77eNDEhJRkK4dvuyD:-My77eNDEhJRkK4dvuyE:b-topnb) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/70f94b2302f0306ef3f63d14c87acb91faa7c621/inspectit-ocelot-documentation/docs/getting-started/installation.md#L96)):*
> ```Markdown
> For further info see [Configuration Sources](../configuration/configuration-sources.md).
> 
> 3. Create an Instrumentation object like the following with spec.java.image set to the inspectIT Ocelot agent image you would like to use (only works with version 1.15.2 and up):
> ```

We might have to update this version number here and in the yaml below depending on which release this change will go live with.

aaronweissler (commented):
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1348)* status: 0 of 4 files reviewed, 2 unresolved discussions

___
*[inspectit-ocelot-agent/docker/entrypoint.sh, line 4 at r3](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1348#-My79wEW5DfL9NcmEkvq:-My79wEW5DfL9NcmEkvr:b-sluvt3) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/70f94b2302f0306ef3f63d14c87acb91faa7c621/inspectit-ocelot-agent/docker/entrypoint.sh#L4)):*
> ```Shell
> # Copy agent jar into shared volume
> cp /inspectit-ocelot-agent.jar /agent/inspectit-ocelot-agent.jar
> while true; do sleep 2; done;
> ```

At least in a docker-compose environment the infinite while loop should not be needed, since containers are only restarted if you set an according restart-policy. But I am not sure if we can/should take it out since at least in theory someone using the image could rely on it?

aaronweissler (commented):
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1348)* status: 0 of 4 files reviewed, 2 unresolved discussions

___
*[inspectit-ocelot-agent/docker/entrypoint.sh, line 4 at r3](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1348#-My79wEW5DfL9NcmEkvq:-My7ByEz7RwkxQzsAI3Y:b-wo5o0h) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/70f94b2302f0306ef3f63d14c87acb91faa7c621/inspectit-ocelot-agent/docker/entrypoint.sh#L4)):*

At least in a docker-compose environment the infinite while loop should not be needed, since containers are only restarted if you set an according restart-policy. But I am not sure if we can/should take it out since at least in theory someone using the image could rely on it?
</blockquote></details>

Especially if someone is already using it in Kubernetes, since there the default value for `restartPolicy` is "Always" and they would have to update it to "OnFailure" or "Never", which might not always be possible/wanted if the container with the agent image is in a Pod with another container since that `restartPolicy` applies to all containers in a Pod.

mariusoe (approved):
> Reviewed 2 of 4 files at r2, 1 of 1 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1348)* status: all files reviewed, 2 unresolved discussions (waiting on @mariusoe)

mariusoe (approved):
> Reviewed 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1348)* status: all files reviewed, 2 unresolved discussions (waiting on @mariusoe)

PR #1376, opened 2022-03-30
fix(config-ui) - Small grammatical corrections in Scope Wizard

aaronweissler:


components/inspectit-ocelot-configurationserver-ui/src/components/views/configuration/scope-wizard/TypeMatcher.js
```diff
@@ -68,7 +68,7 @@ const TypeMatcher = ({ typeMatcher, onTypeMatcherChange, onShowClassBrowser }) =
         <div className="row-center">
           <Dropdown style={{ width: '7rem' }} value={typeMatcher.type} options={types} onChange={(e) => setState('type', e.value)} />
           <label className="inner-label" htmlFor="which">
-            which name
+            name
```
aaronweissler, line 71: I personally preferred simply removing the "which", since it still expresses what it is supposed to just fine and then it is just a bit less somewhat redundant text in the UI. I'd also agree with replacing it with a "whose" which would be analogous to the way it is written in the overview after finishing creating a scope with the scope wizard :)

MariusBrill (commented):
> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1376)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @MariusBrill)

MariusBrill (approved):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1376)* status: all files reviewed, 1 unresolved discussion (waiting on @MariusBrill)

PR #1393, opened 2022-04-13
Closes #1374 - Add Agent Command to retrieve Jar Dependencies

### LINKED ISSUE #1374: [Feature] - Add Agent Command to retrieve Jar Dependencies
**Is your feature request related to a problem? Please describe.**
Add a new Agent Command that allows the user to retrieve the jar dependencies of the application that an inspectIT agent is attached to.

**Describe the solution you'd like**
An agent command should return a formatted list of jar dependencies with the name and version.

**Additional context**
The list of dependencies can be used for several benefits e.g. search for vulnerabilities on the dependency on [CVE Mitre](https://cve.mitre.org/cve/search_cve_list.html) or create a Software Bill of Materials (SBOM)

Mahir-Isikli:
closes #1374 

This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1393)

### DIFF
--- a/components/inspectit-ocelot-configurationserver/src/main/java/rocks/inspectit/ocelot/agentcommunication/handlers/impl/ListDependenciesCommandHandler.java
+++ b/components/inspectit-ocelot-configurationserver/src/main/java/rocks/inspectit/ocelot/agentcommunication/handlers/impl/ListDependenciesCommandHandler.java
@@ -9,43 +9,43 @@ import org.springframework.web.context.request.async.DeferredResult;
 import rocks.inspectit.ocelot.agentcommunication.handlers.CommandHandler;
 import rocks.inspectit.ocelot.commons.models.command.Command;
 import rocks.inspectit.ocelot.commons.models.command.CommandResponse;
-import rocks.inspectit.ocelot.commons.models.command.impl.LogsCommand;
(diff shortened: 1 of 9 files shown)


Mahir-Isikli:
Currently the structure of the new command is done. 

The presently used logic with the 
`RuntimeMXBean bean = ManagementFactory.getRuntimeMXBean();
        String classPath = bean.getClassPath();//` 
is not working as intended and is resulting in an output of only the attached application, i.e. in the case of using the Spring Petclinic.jar the output is 
`path/to/your/app/spring-petclinic.jar:inspectit-ocelot-agent-SNAPSHOT.jar`

danipaniii:
In the related Issue [https://github.com/inspectIT/inspectit-ocelot/issues/1374](url) it was mentioned that with an agent command the dependencies of a jar should be listed. 
After some research I found a partial solution for Maven Projects, where the file structure of the .jar makes it possible to read the dependencies of that jar. However for Gradle Project .jars this method would not work, since they build the Jar differently.

Still there are some tools like jdeps that can be used to get the dependencies of a jar. While I did some personal testing with it, the results where more or less satisfying, which could also be because I don't have much experience with jdeps.

quandor:
If we ever pick this up, we start from scratch.

PR #1424, opened 2022-05-09
Improve exception message for invalid fallback config

### LINKED ISSUE #1418: [Feature] - Test for validation errors in fallback config
**Is your feature request related to a problem? Please describe.**
This is only a problem when modifying the default or fallback configuration, so it should only be a problem for developers, but I think it would be worth fixing for a better development experience.
When putting values in the default config that lead to validation errors, a NoSuchElementException is thrown and its full stack-trace printed, burying the previously logged error-message with the needed information what went wrong exactly.

**Describe the solution you'd like**
Add an ifPresent to check if the fallback configuration could be loaded correctly and log a simple error message if not.

aaronweissler:
Closes #1418

---
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1424)

### DIFF
--- a/inspectit-ocelot-core/src/main/java/rocks/inspectit/ocelot/core/config/InspectitEnvironment.java
+++ b/inspectit-ocelot-core/src/main/java/rocks/inspectit/ocelot/core/config/InspectitEnvironment.java
@@ -215,7 +215,7 @@ public class InspectitEnvironment extends StandardEnvironment {
             if (fallbackConfig.isPresent()) {
                 currentConfig = fallbackConfig.get();
             } else {
-                throw new RuntimeException("The fallback configuration could not be parsed, probably due to a validation error.");
+                throw new RuntimeException("The fallback configuration could not be parsed, probably due to a validation error. See logs for further information.");
             }
 
             fallbackSources.forEach(ps -> propsList.remove(ps.getName()));
--- a/inspectit-ocelot-core/src/main/java/rocks/inspectit/ocelot/core/config/InspectitEnvironment.java
+++ b/inspectit-ocelot-core/src/main/java/rocks/inspectit/ocelot/core/config/InspectitEnvironment.java
@@ -211,7 +211,12 @@ public class InspectitEnvironment extends StandardEnvironment {
             currentSources.forEach(ps -> propsList.remove(ps.getName()));
             fallbackSources.forEach(ps -> propsList.addLast(ps));
 
-            currentConfig = loadAndValidateFromProperties(INSPECTIT_ROOT_PREFIX, InspectitConfig.class).get();
+            Optional<InspectitConfig> fallbackConfig = loadAndValidateFromProperties(INSPECTIT_ROOT_PREFIX, InspectitConfig.class);
+            if (fallbackConfig.isPresent()) {
+                currentConfig = fallbackConfig.get();
+            } else {
+                throw new RuntimeException("The fallback configuration could not be parsed, probably due to a validation error.");
+            }
 
             fallbackSources.forEach(ps -> propsList.remove(ps.getName()));
             currentSources.forEach(ps -> propsList.addLast(ps));


heiko-holz (approved):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1424)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aaronweissler)

___
*[`inspectit-ocelot-core/src/main/java/rocks/inspectit/ocelot/core/config/InspectitEnvironment.java` line 214 at r1](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1424#-N1cItlT8_sa1l_ITDYO:-N1cItlT8_sa1l_ITDYP:bu3zo1g) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/51d27c136a835feed8da8716335e0987086579f5/inspectit-ocelot-core/src/main/java/rocks/inspectit/ocelot/core/config/InspectitEnvironment.java#L214)):*
> ```Java
>             fallbackSources.forEach(ps -> propsList.addLast(ps));
> 
>             Optional<InspectitConfig> fallbackConfig = loadAndValidateFromProperties(INSPECTIT_ROOT_PREFIX, InspectitConfig.class);
> ```

Comment not related to your PR:

I find the code in the `protected void configurePropertySources(Optional<String> cmdLineArgs)` quite confusing.

At firts glane, I cannot see why we expect different behaviour in the `else` block, when we load the properties again. I can only assume that changing the `propsList` has an effect on what `loadAndValidateFromProperties` does. But in general, it is very hard to understand due to missing comments.

If you know what is actually happening, I would actually appreciate one comment why we expect the new config to behave differently than the `initialConfig`

aaronweissler (commented):
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1424)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aaronweissler)

___
*[`inspectit-ocelot-core/src/main/java/rocks/inspectit/ocelot/core/config/InspectitEnvironment.java` line 214 at r1](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1424#-N1cItlT8_sa1l_ITDYO:-N1cOQTpEm-2BcrOIPiA:b-ao8upi) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/51d27c136a835feed8da8716335e0987086579f5/inspectit-ocelot-core/src/main/java/rocks/inspectit/ocelot/core/config/InspectitEnvironment.java#L214)):*

Comment not related to your PR:

I find the code in the `protected void configurePropertySources(Optional<String> cmdLineArgs)` quite confusing.

At firts glane, I cannot see why we expect different behaviour in the `else` block, when we load the properties again. I can only assume that changing the `propsList` has an effect on what `loadAndValidateFromProperties` does. But in general, it is very hard to understand due to missing comments.

If you know what is actually happening, I would actually appreciate one comment why we expect the new config to behave differently than the `initialConfig`
</blockquote></details>

It definitely is not that easy to understand at a glance, I agree.

`propsList` are the actual PropertySources of the InspectitEnvironment in mutable form (line 176), so changing `propsList` actually changes the Properties that are loaded in `loadAndValidateFromProperties` in line 252 with `ConfigurationPropertySources.get(this)`. 

And they are different if the  `currentSources` are different to the `fallbackSources`, which would be the case if any user-defined configuration had been added. 
Now if the error is in the default configuration, the behaviour will not be different and the same parsing error will be logged twice, because the invalid value will be in both the `currentSources` and `fallbackSources`, because both always include the default config.

heiko-holz (approved):
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1424)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aaronweissler)

PR #1481, opened 2022-06-20
Test for vulnerabilities in Java dependencies with Snyk

aaronweissler:
Still needs an environment `snyk` with a Snyk Auth-Token set as secret `SNYK_AUTH`.

When setting that up, we can also set up checking individual PRs for new vulnerabilities.

---
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1481)


heiko-holz (approved):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1481)* status: all files reviewed, 1 unresolved discussion (waiting on @aaronweissler)

___
*[`.github/workflows/snyk_test_sarif.yml` line 40 at r1](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1481#-NBuYd6Q2rVTM1qOySFc:-NBuYd6Q2rVTM1qOySFd:b-sf91qd) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/5406b6af8d03ec0f505831896353ed76f9793911/.github/workflows/snyk_test_sarif.yml#L40)):*
> ```YAML
>         continue-on-error: true
>       - name: Upload SARIF file
>         if: steps.test.outcome == 'success'
> ```

where do we get that value from?
How does the `outcome` differ from the `exit_code` in the file `snyk-test-script.sh`?

aaronweissler (commented):
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1481)* status: all files reviewed, 1 unresolved discussion (waiting on @heiko-holz)

___
*[`.github/workflows/snyk_test_sarif.yml` line 40 at r1](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1481#-NBuYd6Q2rVTM1qOySFc:-NBvY0YJECkDDmbouAoY:b6gx3p7) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/5406b6af8d03ec0f505831896353ed76f9793911/.github/workflows/snyk_test_sarif.yml#L40)):*

where do we get that value from?
How does the `outcome` differ from the `exit_code` in the file `snyk-test-script.sh`?
</blockquote></details>

The outcome just says if any of the commands in the step failed, so in this case it is basically just a mapping of the exit code of `snyk-test-script.sh` itself where `0` is mapped to `success`and `1` is mapped to `failure`.
So it differs from the values of the `exit_code` variable in the way, that if `exit_code` is `1` the shell script's exit code is still `0` because we still want the SARIF file to be uploaded, since it contains the information on the found errors.
The if condition and therefore that abstraction is needed, because `snyk test` returns exit code `2` when the command simply fails, and then we don't want to upload the SARIF, because it would be empty and all previously reported errors would be marked as resolved by GitHub, since the uploaded SARIF file was empty.

heiko-holz (approved):
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1481)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aaronweissler)

PR #1488, opened 2022-06-27
Highlight enums with String values correctly

### LINKED ISSUE #1436: [Bug] - Config Server falsely marks grpc as a wrong protocol label.
**Describe the bug**
The config server marks the value `grpc` in `inspectit.exporters.tracing.jaeger. protocol` as wrong. However, the value is correct.

**Expected behavior**
The value should not be marked as an error.

aaronweissler:
Closes #1436

---
This change is [Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1488)

### DIFF
--- a/components/inspectit-ocelot-configurationserver/src/main/java/rocks/inspectit/ocelot/rest/yamlhighlighter/HighlightRulesMapController.java
+++ b/components/inspectit-ocelot-configurationserver/src/main/java/rocks/inspectit/ocelot/rest/yamlhighlighter/HighlightRulesMapController.java
@@ -3,24 +3,24 @@ package rocks.inspectit.ocelot.rest.yamlhighlighter;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.CaseFormat;
 import io.swagger.annotations.ApiOperation;
+import lombok.extern.slf4j.Slf4j;
 import org.springframework.http.MediaType;
 import org.springframework.util.ReflectionUtils;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RestController;
 import rocks.inspectit.ocelot.config.model.InspectitConfig;
+import rocks.inspectit.ocelot.config.model.exporters.TransportProtocol;
 import rocks.inspectit.ocelot.config.model.instrumentation.actions.GenericActionSettings;
 import rocks.inspectit.ocelot.rest.AbstractBaseController;
 
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.ParameterizedType;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.stream.Collectors;
 
 @RestController
+@Slf4j
 public class HighlightRulesMapController extends AbstractBaseController {
 
     @VisibleForTesting
@@ -166,7 +166,20 @@ public class HighlightRulesMapController extends AbstractBaseController {
      * @return List of names of all possible values for the enum.
      */
     private List<String> extractEnumValues(Class<?> currentEnum) {
-        return Arrays.stream(currentEnum.getFields()).map(Field::getName).collect(Collectors.toList());
+        List<String> allFields = Arrays.stream(currentEnum.getDeclaredFields()).map(Field::getName).collect(Collectors.toList());
+        if (allFields.contains("values")){
+            try {
+                Field valuesField = currentEnum.getDeclaredField("values");
+                valuesField.setAccessible(true);
+                Map<String, ?> valuesMap = (Map<String, ?>) valuesField.get(null);
+                return new ArrayList<>(valuesMap.keySet());
+            } catch(Exception e) {
+                log.error(String.format("Error while getting values of enum %s.", currentEnum.getName()), e);
+                return Collections.emptyList();
+            }
+        } else {
+            return Arrays.stream(currentEnum.getFields()).map(Field::getName).collect(Collectors.toList());
+        }
     }
 
     @ApiOperation(value = "Get JSON for Highlight Rules Generation", notes = "")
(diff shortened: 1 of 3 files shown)


heiko-holz (changes requested):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1488)* status: all files reviewed, 1 unresolved discussion (waiting on @aaronweissler)

___
*[`components/inspectit-ocelot-configurationserver/src/main/java/rocks/inspectit/ocelot/rest/yamlhighlighter/HighlightRulesMapController.java` line 170 at r1](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1488#-N6gnQfw8YgZP0FbVgfj:-N6gnQfw8YgZP0FbVgfk:b34fuj1) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/2930f3e7be9b5ec22787cf13ce3ffa3ec8ee9e29/components/inspectit-ocelot-configurationserver/src/main/java/rocks/inspectit/ocelot/rest/yamlhighlighter/HighlightRulesMapController.java#L170)):*
> ```Java
>     private List<String> extractEnumValues(Class<?> currentEnum) {
>         List<String> allFields = Arrays.stream(currentEnum.getDeclaredFields()).map(Field::getName).collect(Collectors.toList());
>         if (allFields.contains("values")){
> ```

The field `values` is specific to the `TransportProtocol`.

In case of the `TransportProtocol` it would also be sufficient to just map to the `toString` method instead of the `getName`. 

Assuming that the other Enums do not override the `toString` method to express something different, we can use

```
 private List<String> extractEnumValues(Class<?> currentEnum) {
        return Arrays.stream(currentEnum.getEnumConstants()).map(Object::toString).collect(Collectors.toList());
    }
``

(in order to make the test pass, the order of the expected values for `TransportProtocol` need to be adjusted in `HighlightRulesMapControllerIntTest`

Alternatively, we can also add the edge case of the `TransportProcotol` directly into the `extractEnumValues` method:

```
 private List<String> extractEnumValues(Class<?> currentEnum) {
        Class<? extends Enum<?>> clazz = (Class<? extends Enum<?>>) currentEnum;
        if (currentEnum.equals(TransportProtocol.class)) {
            return Arrays.stream(clazz.getEnumConstants()).map(Enum::toString).collect(Collectors.toList());
        } else {
            return Arrays.stream(clazz.getEnumConstants()).map(Enum::name).collect(Collectors.toList());
        }
    }
```
```

aaronweissler (commented):
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1488)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @heiko-holz)

___
*[`components/inspectit-ocelot-configurationserver/src/main/java/rocks/inspectit/ocelot/rest/yamlhighlighter/HighlightRulesMapController.java` line 170 at r1](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1488#-N6gnQfw8YgZP0FbVgfj:-N6i2cKf82kWD2bsIJUq:b83gc9x) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/2930f3e7be9b5ec22787cf13ce3ffa3ec8ee9e29/components/inspectit-ocelot-configurationserver/src/main/java/rocks/inspectit/ocelot/rest/yamlhighlighter/HighlightRulesMapController.java#L170)):*

The field `values` is specific to the `TransportProtocol`.

In case of the `TransportProtocol` it would also be sufficient to just map to the `toString` method instead of the `getName`. 

Assuming that the other Enums do not override the `toString` method to express something different, we can use

```
 private List<String> extractEnumValues(Class<?> currentEnum) {
        return Arrays.stream(currentEnum.getEnumConstants()).map(Object::toString).collect(Collectors.toList());
    }
``

(in order to make the test pass, the order of the expected values for `TransportProtocol` need to be adjusted in `HighlightRulesMapControllerIntTest`

Alternatively, we can also add the edge case of the `TransportProcotol` directly into the `extractEnumValues` method:

```
 private List<String> extractEnumValues(Class<?> currentEnum) {
        Class<? extends Enum<?>> clazz = (Class<? extends Enum<?>>) currentEnum;
        if (currentEnum.equals(TransportProtocol.class)) {
            return Arrays.stream(clazz.getEnumConstants()).map(Enum::toString).collect(Collectors.toList());
        } else {
            return Arrays.stream(clazz.getEnumConstants()).map(Enum::name).collect(Collectors.toList());
        }
    }
```
```

</blockquote></details>

Yes, exactly this problem/limitation I would have mentioned when turning it from draft to reviewable, that right now it assumes that any other Enums using this functionality would also save their values in a `values` field.

Simply using toString seems like a really good idea to me though, since it should be correct for any Enums used in this context, since the values for configurations are also defined as Strings. 
And I like it more than defining an exception specifically for TransportProtocol, since it would just work for new Enums working the same way without additional code changes.

heiko-holz (approved):
> Reviewed 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1488)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aaronweissler)

___
*[`components/inspectit-ocelot-configurationserver/src/main/java/rocks/inspectit/ocelot/rest/yamlhighlighter/HighlightRulesMapController.java` line 170 at r1](https://reviewable.io/reviews/inspectit/inspectit-ocelot/1488#-N6gnQfw8YgZP0FbVgfj:-N6lCkl5B5BxeDO_6Ppp:b-vs6d3r) ([raw file](https://github.com/inspectit/inspectit-ocelot/blob/2930f3e7be9b5ec22787cf13ce3ffa3ec8ee9e29/components/inspectit-ocelot-configurationserver/src/main/java/rocks/inspectit/ocelot/rest/yamlhighlighter/HighlightRulesMapController.java#L170)):*

Yes, exactly this problem/limitation I would have mentioned when turning it from draft to reviewable, that right now it assumes that any other Enums using this functionality would also save their values in a `values` field.

Simply using toString seems like a really good idea to me though, since it should be correct for any Enums used in this context, since the values for configurations are also defined as Strings. 
And I like it more than defining an exception specifically for TransportProtocol, since it would just work for new Enums working the same way without additional code changes.
</blockquote></details>

Yep, I also like the implementation with `toString`

PR #1576, opened 2023-02-13
Updates dependency of Agent to recent versions

quandor:
Most of the dependencies are now managed via spring-boot dependency management. That means an upgrade of spring-boot version updates automatically dependencies managed by spring-boot *and* guarantees compatibility.
All other versions are managed in a single file (`gradle.properties`). Usually those versions are the most current ones. If that is not the case a comment should explain why not.

- - -
This change is [Reviewable](https://reviewable.io/reviews/inspectIT/inspectit-ocelot/1576)


inspectit-ocelot-agent/src/system-test/java/rocks/inspectit/ocelot/instrumentation/special/ExecutorContextPropagationTest.java
```diff
@@ -62,7 +62,7 @@ public void correlateRunnable_lambda() throws Exception {
 
             taskFuture.get();
 
-            assertThat(refTags.get()).hasSize(1)
+            assertThat(refTags.get()).toIterable().hasSize(1)
```
quandor, line 65: That change was necessary to be able to use a current version of `assertj-core`.

inspectit-ocelot-core/src/test/java/rocks/inspectit/ocelot/core/exporter/TransportProtocolSubsetValidatorTest.java
```diff
@@ -27,7 +27,7 @@ public class TransportProtocolSubsetValidatorTest {
     private Validator validator;
 
     @BeforeEach
-    private void beforeEach() {
```
quandor, line 30: That change was necessary to be able to use current versions of JUnit.

.github/workflows/agent_test.yml
```diff
@@ -24,13 +24,10 @@ jobs:
       fail-fast: false
       matrix:
         dockerimage:
-          - 'openjdk:8-jdk'
```
quandor, line 27: I included the following information in the corresponding commit comment. That ensures that reasons for this decision are not completely lost in a distant future.

Why did I change those images?
* The images of sgrio are at least two years old.
  See https://hub.docker.com/r/sgrio/java
* Images from openjdk are deprecated.
  See https://hub.docker.com/_/openjdk

eclipse-temurin are the official openjdk builds. Since they are based on
ubuntu (as far as the comitter understands it) there is no reason for
ubuntu specific images in the build matrix anymore.

PR #1588, opened 2023-03-15
Closes #1587 - Fix Chanelog generation in Github Release Action

### LINKED ISSUE #1587: [Feature] - Fix changelog generation for Github Release Actoin
Now the changelog gets generated twice by  two different action, but we want to just have the github changelog generator to generate the changelog.

danipaniii:
Closes #1587


quandor (approved):
Sounds like a good idea to not generate release note, if a previous step already generates them.

PR #1643, opened 2024-01-23
Update README.md

SaschaBrechmannVHV:
Change OpenCensus Lib to OpenTelemetry SDK

- - -
This change is [Reviewable](https://reviewable.io/reviews/inspectIT/inspectit-ocelot/1643)


EddeCCC (changes requested):
@SaschaBrechmannVHV Currently we use both OpenCensus and OpenTelemetry, thus we should mention both of them.

SaschaBrechmannVHV:
Hi @EddeCCC ,
are you shure that we use both, OT and OC? I think with the 2.0 Release we migrated from OC to OT and use the OC Shim from OT?
Regards, Sascha

EddeCCC:
Hi @SaschaBrechmannVHV,

yes we also use the OC Shim together with OT. However, we still need the OC SDK for the metric exporters, as the shim only inludes opencensus-impl-core.
Thus, we currently use both.

Regards, Edi

SaschaBrechmannVHV:
@EddeCCC : I updated the Links

EddeCCC (approved):

PR #1669, opened 2024-11-08
fix ui-bug

EddeCCC:
The status page might render too early, resulting in a null error. This PR adds a null check for this situation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/inspectIT/inspectit-ocelot/1669)


EddeCCC:
Also include Dependabot bumps

PR #1673, opened 2024-11-11
fix: Use Testcontainer for InfluxExporterIntTest 

levinkerschberger:
There was a problem with the download of InfluxDB, when using JDK 8 from IBM.  Should be fixed with testcontainer setup.

- - -
This change is [Reviewable](https://reviewable.io/reviews/inspectIT/inspectit-ocelot/1673)


EddeCCC (approved):
Awesome

Issue #1705, opened 2025-04-15
[Feature] - Recycle temporary JAR files


EddeCCC:
**Is your feature request related to a problem? Please describe.**
When running multiple agents on the same machine, every agent creates three temporary JAR files, which are needed for runtime instrumentation. If all the agents have the same version, this uses an unnecessary amount of space on the system.

**Describe the solution you'd like**
Agents should have an option to recycle such files. If an agent has started in recycle mode, it will look for such JAR files. If no files are found, they will be created (and not deleted after shutdown).

**Additional context**
- It should also be possible to set the path, where such temporary files will be created.
- It should be possible to run different agent versions on the same machine without conflicts.

PR #1711, opened 2025-04-16
Add recycling of jars

EddeCCC:
Closes https://github.com/inspectIT/inspectit-ocelot/issues/1705


EddeCCC:
There are `ClassNotFoundException`s, when starting multiple agent at the same time with enabled recycling. Probably, because the files created by the first agent are not "ready" yet.
Using a `inspectit.start.delay` of 100 ms works, but is this the solution? Another way would be to somehow prepare the files beforehand.

PR #1740, opened 2025-05-14
Update dependencies

EddeCCC:


EddeCCC:
For some reason, the equals() methods of (some) ByteBuddy element matchers behave differently since 1.15.5
Thus, comparing matchers in tests leads to failing, even tough we expect them to be equal.

EddeCCC:
See: https://github.com/raphw/byte-buddy/issues/1813

PR #1839, opened 2025-08-28
Update dependencies

EddeCCC:


EddeCCC:
DependencyCheck-plugin has trouble connecting to database

